PR #1, opened 2017-05-30
Mount in the agent binary

lox:
This mounts the buildkite-agent binary into the container from the host and provides the minimum env.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 A [Buildkite](https://buildkite.com/) plugin for running pipeline steps in [Docker](https://www.docker.com/) containers
 
+The `buildkite-agent` command line tool (and required environment variables) will also be mounted into the container, allowing you to use it for artifact download, etc.
+
 If you need more control, please see the [docker-compose Buildkite Plugin](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin).
 
 The docker container has the host buildkite-agent binary mounted in to `/usr/bin/buildkite-agent` and the required environment variables set. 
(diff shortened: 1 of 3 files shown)


lox:
I added a config key for the which agent binary to use, if you pass `false` it will disable the behaviour entirely.

toolmantim:
Whilst trying to rework the docs for the option, a few questions came up…

* Should it be called `buildkite-agent-path`?
* If my image has a custom `buildkite-agent` binary, then I'd still need the environment variables (which `"false"` would disable)?
* Is there really a case for people to set a custom path (something other than `"false"`). I imagine it's either "Yeah, automatically mount it" or "No please don't mount it, we have our own binary"

What I landed on was this:

```
### `mount-buildkite-agent` (optional)

Whether to automatically mount the `buildkite-agent` binary from the host agent machine into the container. Defaults to `"true"`. Set to `"false"` if you want to disable, or if you already have your own binary in the image.
```

And to always pass in the environment variables.

What about the follow for the top paragraph (above "If you need more control")?

```
The `buildkite-agent` command line tool (and required environment variables) will also be mounted into the container, allowing you to use it for artifact download, etc.
```

Also do you know what we can't use YAML `false/true`? Is that something we could fix in our backend or agent, or just something we can't support because YAML?

lox:
Yup, think that makes a heap of sense @toolmantim. 

I was actually thinking of a slightly different problem with YAML, which is that strings like "on" translate too booleans. Good catch :)

http://yaml-online-parser.appspot.com/?yaml=---%0Aenable_llamas%3A+%5B+true%2C+%22true%22%2C+on+%5D%0A%0A&type=json

lox:
Made those changes @toolmantim.

PR #6, opened 2017-10-19
Add user and shell options

mikeknox:
PR to allow a shell or user to be specified to the docker run.
Previously defaulted to bash, which doesn't ship by default in Alpine.

Also added a build option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 A [Buildkite](https://buildkite.com/) plugin for running pipeline steps in [Docker](https://www.docker.com/) containers
 
+The `buildkite-agent` command line tool (and required environment variables) will also be mounted into the container, allowing you to use it for artifact download, etc.
+
 If you need more control, please see the [docker-compose Buildkite Plugin](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin).
 
 The docker container has the host buildkite-agent binary mounted in to `/usr/bin/buildkite-agent` and the required environment variables set. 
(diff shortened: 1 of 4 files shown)


toolmantim:
Thanks for taking the time for putting together such a thorough PR!

I have some questions though. I thought the `build` option might be able to used with the `run` option, as a way of doing a build + run type thing. But that doesn't appear to be the case? What's the use case for just a build? Buildkite steps are designed to be stateless, so just wondering what you do with the image after it's built?

toolmantim:
Also, thanks for your patience waiting for a review too! 🙇🏼‍♂️

lox:
I totally missed this! Will check it out and come back with some feedback soon. Sorry for taking so long @mikeknox!

lox:
I'd be tempted to use the docker-compose syntax https://docs.docker.com/compose/compose-file/#build:

```yaml
build:
  context: ./dir
  dockerfile: Dockerfile-alternate
```

or the simpler alternative:

```
build: ./dir
```

I assumed the same as @tim, build is just an alternative to `image:`, right? It should still then run a command in that image.

mikeknox:
Thanks Tim, Lachlan,
After working with this, I'm inclined to drop the build addition from my PR
as the 'docker-compose' plugin provides more comprehensive coverage of the
build+run use-case.
The purpose of the 'build: true' was to just build the container, not to
run it, as I was working through a scenario where I just needed to
build+push the container, not run it.

I'd like go forward with the additional parameters though such as 'shell'
and 'user'.

toolmantim:
Okay, no worries at all. Would love to merge the `shell` and `user` options in.

mikeknox:
I've now updated this PR to remove the `build` option and its artefacts.

toolmantim:
Thanks @maktouch, I've just merged the other PRs. Are you thinking of splitting this up one too? (and adding the test for the `shell` option?) If not, I could help.

mikeknox:
I've added the test for `shell` to PR#6

toolmantim:
Thanks again @mikeknox 🎉

lox:
I'm just in the process of cutting a release of v1.1.0 (belatedly).

I've decided to drop the `shell` option. Apologies for not providing more feedback whilst it was happening, I initially assumed it was a `docker run` param like `--user`, but it looks like it's a way to specify a shell binary that is used to invoke `$BUILDKITE_COMMAND`. I'm assuming this was done to support multi-line comments. 

IMO this is surprising and not docker-like, and beyond that it breaks for any shell that doesn't have `-c`, `cmd.exe` for instance. 

See https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/59 for some more backstory on multiple commands.

mikeknox:
fair enough, the shell option was an extension of the original
implementation which had `bash` hard-coded as the shell.
Dropping that hard-coding and the shell option makes sense, and a shell can
be provided via the command if required.

PR #8, opened 2017-12-02
Make workdir optional and set a sensible default.

maktouch:
This patch makes workdir optional and adds a default.

In a docker pipeline build, the workdir is most of the time irrelevant as long as the command runs in the same path as the code.

Less required stuff, easier adoption :-)


toolmantim:
Thanks @maktouch!

PR #9, opened 2017-12-02
Added option to always pull image before running command

maktouch:
Added readme, tests, and the actual logic.


toolmantim:
Thanks @maktouch!

PR #10, opened 2017-12-02
Added possibility to mounts extra volumes

maktouch:
This patch adds the possibility to mounts extra volumes.

Once #6, #8 and #9 are merged, it should cover all docker primitives (people will no longer need to fork this plugin to make their stuff work, pretty much everything would be possible from the options).


toolmantim:
Thanks @maktouch! This is great.

PR #12, opened 2018-01-24
Add back "bash -c" before commands

lox:
Turns out I broke backwards compatibility with 1.0.0 with the 1.1.0 release. I missed that we'd used `bash -c` as a command runner for all commands. This adds it back, and we will drop it shortly in a 2.0.0 release. 

Apologies folks!

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -59,8 +59,4 @@ if [[ "${debug_mode:-off}" =~ (on) ]] ; then
     "${BUILDKITE_COMMAND}" >&2
 fi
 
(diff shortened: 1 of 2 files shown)


thomasbarton:
@lox thank you for looking into this 👍

toolmantim:
I'll put out a release now

toolmantim:
v1.1.1 has this fix in. Sorry about that @thomasbarton!

https://github.com/buildkite-plugins/docker-buildkite-plugin/releases/tag/v1.1.1

Issue #13, opened 2018-01-31
Feature Request: Add support for specifying docker network


roblugton:
Hi there,

I would like to be able to connect a build container to a 'custom' docker network when using the plugin. 

Ideally it should create the network if it does not already exist.

Will have a play around and submit a pull request later today.

Cheers

Rob

lox:
Would that just be a matter of exposing `network` from docker run? https://docs.docker.com/engine/reference/run/#network-settings 

I wonder if we should just investigate a generic way of exposing all of the docker run parameters. Thoughts @toolmantim?

toolmantim:
Hrmm… or we just add all the missing options? Could be a helluvalot of options though! Maybe a `docker-args` plugin property where it just passes through?

lox:
There are 98 different run options @toolmantim :) 

I kind of like the idea of an `args` property that just passes straight through. 

In terms of the "create a network before running", I reckon that one starts to get into `docker-compose` territory @roblugton. Most folks end up getting to where you are and then deciding it's easier to create an docker-compose.yml!

roblugton:
It certainly does :) The use case is I'm using the docker plugin to build something that spins up a docker-compose stack as part of it. The 'simplest' way i've found to get the build container to talk to the containers spun up is to join them all to the same network (this is using sibling-containers btw) 

Here's what I've got so far. The test is still failing for some reason (i've no experience wit bats! 🦇 ) so once I've fixed that I'll submit a pull request: https://github.com/roblugton/docker-buildkite-plugin/commit/918bb93646c3d01db0a174d1a52f2e641de14e3b

lox:
If you uncomment the DOCKER_STUB_DEBUG line at the top, you get debugging. Unfortunately for all tests, but if you look at the debug from before the test:

```
 bats-mock(docker): got docker network ls --quiet --filter name=foo                                                                                                                                         8/9
bats-mock(docker): match failed at idx 1, expected 'create', got 'ls'
bats-mock(docker): result 0
bats-mock(docker): got docker network create foo
bats-mock(docker): match failed at idx 0, expected 'run', got 'network'
bats-mock(docker): result 0
bats-mock(docker): got docker run -it --rm --volume /Users/lachlan/Projects/buildkite/roblugton-docker-buildkite-plugin:/app --workdir /app --network foo image:tag bash -c echo hello world
bats-mock(docker): no plan row found
bats-mock(docker): result 0
```

[...]

PR #14, opened 2018-02-01
Add support for connecting the container to a docker network

roblugton:
Enables connecting the docker container to the specified docker network. Will create the network if it does not exist. 

(See issue #13 for more info)


lox:
Perfection! Thanks so much for this! 🙌🏻

PR #16, opened 2018-03-13
Add 'srcdir' plugin option

tduffield:
This value allows you to specify which directory (relative to $PWD) you wish to mount as the `workdir` in your container. This is designed to be used with Docker Images that have hard-coded ENTRYPOINTs.

### Example
```
steps:
  - label: terraform
    commands:
      - fmt -check=true
      - validate -check-variables=false
    plugins:
     docker#v1.1.X:
        image: "hashicorp/terraform"
        srcdir: "./terraform"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,12 @@ The working directory where the pipeline’s code will be mounted to, and run fr
 
 Example: `/app`
 
+### `srcdir` (optional)
+
+The relative path to the source directory you wish to mount into the running container. The default is `./`.
+
+Example: `./components/my-source`
+
 ### `always-pull` (optional)
 
 Whether to always pull the latest image before running the command. Useful if the image has a `latest` tag. The default is false, the image will only get pulled if not present.
@@ -99,7 +105,7 @@ Example: `root`
 
 ### `network` (optional)
 
-Join the container to the docker network specified. The network will be created if it does not already exist. See https://docs.docker.com/engine/reference/run/#network-settings for more details. 
+Join the container to the docker network specified. The network will be created if it does not already exist. See https://docs.docker.com/engine/reference/run/#network-settings for more details.
 
 Example: `test-network`
 
--- a/hooks/command
+++ b/hooks/command
@@ -9,10 +9,12 @@ if [[ "${BUILDKITE_PLUGIN_DOCKER_DEBUG:-false}" =~ (true|on|1) ]] ; then
   set -x
 fi
 
+srcdir=$(realpath $PWD/${BUILDKITE_PLUGIN_DOCKER_SRCDIR:-./})
+
 args=(
   "-it"
   "--rm"
-  "--volume" "$PWD:${BUILDKITE_PLUGIN_DOCKER_WORKDIR:-/workdir}"
+  "--volume" "$srcdir:${BUILDKITE_PLUGIN_DOCKER_WORKDIR:-/workdir}"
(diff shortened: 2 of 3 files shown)


lox:
Thanks for this! I wonder if there is some other way we could accomplish this without adding it to the plugin.

lox:
I think setting `workdir` should work:

```
steps:
  - label: terraform
    commands:
      - fmt -check=true
      - validate -check-variables=false
    plugins:
     docker#v1.1.X:
        image: "hashicorp/terraform"
        workdir: "/workdir/terraform"
```

tduffield:
@lox so, my reading of the code is that the host-system path being mounted into the container is always `$PWD` (e.g. the root of the project): https://github.com/buildkite-plugins/docker-buildkite-plugin/blob/master/hooks/command#L15. So even if we set the work directory, the contents of `/workdir/terraform` would still be the root of my directory.

lox:
$PWD will mount to `/workdir` in the container and the container gets a working directory of `/workdir`. If you set `workdir` for the plugin, it will keep the same mount dir, but change the working directory in the container. Provided terraform is using the $PWD inside the container, it should all be relative to `/workdir/terraform`.

lox:
Unless I'm misunderstanding! 😓

tduffield:
> The agent parses the plugin block and passes the parameters through to the script as environment variables in the form BUILDKITE_PLUGIN_{NAME}_{KEY}, or if a list is provided then BUILDKITE_PLUGIN_{NAME}_{KEY}_{IDX}.

source: https://buildkite.com/docs/agent/v2/plugins#developing-a-plugin

So let's say you set `workdir` on the `docker` plugin to `/workdir/terraform` like you did in your example above. Based on the way the documentation is written, you are setting `BUILDKITE_PLUGIN_DOCKER_WORKDIR` to `/workdir/terraform`. Given the line I posted above, that would mean we are mounting `$PWD` to `/workdir/terraform`. For most jobs, unless you have specifically modified the CWD via a  hook, `$PWD` will be the root of the repository.

lox:
Urgh, you are totally correct. I mistakenly assumed that workdir corresponded to docker's `workdir` directive, which it doesn't. Let me re-evaluate.

lox:
E.g https://docs.docker.com/engine/reference/run/#workdir

lox:
Ok, so the issue is that we are conflating `workdir` which is the working directory inside the container with the path that we are mounting $PWD onto. 

It's a slightly different take to yours, but what if we exposed `mountdir` for specifying where $PWD is mounted, then `workdir` can behave exactly like it's docker counterpart and not confuse us next time.

lox:
I'm also coming around to `srcdir` as you implemented it :) Otherwise the default `workdir` of `/workdir` becomes confusing.

tduffield:
I think if we do `mountdir` as I understand it, you're breaking backwards "mental compatibility" because you're fundamentally changing how `/workdir` is used. The benefit of `srcdir` is that it maintains that backwards compatibility. 

That being said, it might be better to do that break now, bump the major/minor version, and create a paradigm that is more inline with Docker's nomenclature.

lox:
> That being said, it might be better to do that break now, bump the major/minor version, and create a paradigm that is more inline with Docker's nomenclature.

I agree. We'd really like to keep this plugin as close to how `docker-run` is used as possible.

PR #19, opened 2018-04-15
Allow pwd mount to be skipped

lox:
Kubernetes on Google Container OS doesn't like host mounts, so this adds a `mount-checkout-path` option so that this behaviour can be disabled.

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -16,7 +16,7 @@ args=(
 # Handle the mount-checkout-path option
 if [[ "${BUILDKITE_PLUGIN_DOCKER_MOUNT_CHECKOUT_PATH:-true}" =~ (true|on|1) ]] ; then
   args+=(
-    "--volume" "${BUILDKITE_BUILD_CHECKOUT_PATH}:${BUILDKITE_PLUGIN_DOCKER_WORKDIR:-/workdir}"
+    "--volume" "${PWD}:${BUILDKITE_PLUGIN_DOCKER_WORKDIR:-/workdir}"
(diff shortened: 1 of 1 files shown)


nbering:
I'm looking at this for a way to pre-build a single image and push to a registry, then parallelize all my tests against the image. The image would effectively replace the git commit sha for the rest of the pipeline.

I can use the docker-compose to build and push the initial image, but the "build" and "image" parameters in docker-compose seem incompatible. I'm thinking I either need to generate a new docker-compose field with the appropriate image reference at build time, or use the plain docker plugin for next steps... but the auto-mounted work directory is just in the way for this use case.

nbering:
I don't have access to the build output. What was the failure?

lox:
This could probably be merged now @toolmantim. I can see it being useful when you don't actually want source in the container.

As more time goes on, I regret making the default behaviour that source and agent are mounted in by default.

nbering:
I did figure my case out in the end with docker-compose plugin, but would still be nice to have this. One of my other pipelines occasionally has issues with dangling files with ownership issues that are a pain to clean up, since the buildkite-agent user isn't the owner.

nbering:
I ran into another case where this "would be nice". I just want to run a step in a container without any real ties to the source repository. Basically just a step to download some thinks as attach them as build artifacts for another step - and there's really no need to mount the checkout directory for this. Any word on whether this can be merged?

lox:
I think this is better handled by #69.

PR #20, opened 2018-04-19
Support additional groups (--group-add)

haines:
This PR adds an `additional_groups` option to the plugin which takes an array of groups (either names or ids), and translates them to [`--group-add` options](https://docs.docker.com/engine/reference/run/#additional-groups) to pass to `docker run`.

I wanted to be able to do this because I'm trying to do some docker-in-docker stuff as an unprivileged user, so I need to be in the `docker` group. My config would then be something like

```yaml
docker#v1.1.1:
  image: "docker:latest"

  mounts:
    - /var/run/docker.sock:/var/run/docker.sock
    - # other mounts

  user: "501:502" # uid:gid of the user on the host, for access to the other mounts
  additional_groups:
    - 501 # gid of docker group on the host, for access to the docker socket
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,12 @@ Allows a user to be set, and override the USER entry in the Dockerfile. See http
 
 Example: `root`
 
+### `additional_groups` (optional)
+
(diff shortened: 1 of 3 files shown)


toolmantim:
Thanks for th contribution! The PR looks great.

PR #32, opened 2018-05-04
Improvements

arikkfir:
Several improvements:

- Pass `SOURCE_UID` and `SOURCE_GID` to container as environment variables.
- Fix debugging output.
- Use an alternative method of passing arguments to the Docker `run` command, which supports commands comprised of multiple tokens.
- Support loading custom hostnames to pass to the container from a `/etc/buildkite-agent/hosts` file (passed as `--add-host` argument to the `docker run` command)

We found these improvements to be useful in multiple occasions, specifically:
- the `SOURCE_UID` and `SOURCE_GID` parameters are useful for containers that need to persist some changes in files under the workdir, and ensure they can be read by the outer process (ie. the buildkite agent)
- the alternative method of running the container is meant to ensure that the `BUILDKITE_COMMAND` argument is provided to the `docker run ...` command as multiple arguments (parsed by `bash`) instead of just one string.

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -17,8 +17,8 @@ args=(
 )
 
 # built-in support for the "USER_UID" and "USER_GID" environment variables, initialized to the BuildKite agent's user
-args+=( "--env" "USER_UID=$(id -u)" )
-args+=( "--env" "USER_GID=$(id -g)" )
+args+=( "--env" "SOURCE_UID=$(id -u)" )
+args+=( "--env" "SOURCE_GID=$(id -g)" )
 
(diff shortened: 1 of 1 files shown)


arikkfir:
Please ignore this - I will split this into smaller individual pull requests.

arikkfir:
Hi @toolmantim yeah I'm trying to find the time the break it up to pieces for you to add if you wanted to... unsuccessful so far :sweat_smile: 

But until then please feel free to reuse anything you find in the code, there's nothing really sophisticated there...

toolmantim:
No worries! It's all here for future Google travellers too.

PR #43, opened 2018-06-07
Don't echo env vars

toolmantim:
We print the environment key values from the pipeline.yml to the build log. Should we do this always? Should we do it when `debug: true` is enabled (as it's done when it prints the args)

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -45,7 +45,7 @@ fi
 # Parse extra env vars and add them to the docker args
 while IFS='=' read -r name _ ; do
   if [[ $name =~ ^(BUILDKITE_PLUGIN_DOCKER_ENVIRONMENT_[0-9]+) ]] ; then
-    echo "parsed env[$name] = [${!name}]"
+    echo "parsed env[$name]"
     args+=( "--env" "${!name}" )
   fi
 done < <(env | sort)


hooks/command
```diff
@@ -45,7 +45,7 @@ fi
 # Parse extra env vars and add them to the docker args
 while IFS='=' read -r name _ ; do
   if [[ $name =~ ^(BUILDKITE_PLUGIN_DOCKER_ENVIRONMENT_[0-9]+) ]] ; then
-    echo "parsed env[$name] = [${!name}]"
+    echo "parsed env[$name]"
     args+=( "--env" "${!name}" )
   fi
 done < <(env | sort)
```
sj26: I think this env parser will fail with multi-line values:

```
$ export HELLO="here's
> a
> multiline
> string"
$ env
SHELL=/bin/bash
HELLO=here's
a
multiline
string
PWD=/Users/sj26
EDITOR=vim
```

toolmantim:
I think the echo is just something left around from testing, and we can safely remove it. It's printed later when the full command is output.

I don't think this whole thing is a problem security-wise though. If someone sets `debug: true`, and uploads their pipeline.yml with an secret env interpolated into the yml, it'll be in the build output. But that's not a good thing anyone should be doing anyhow.

hooks/command
lox: I don't think so @sj26, the loop is just to find the environment keys, then the keys are looked up from the environment. There should have been a better comment story in there to explain that. What this will be vulnerable to is values with `BUILDKITE_PLUGIN_DOCKER_ENVIRONMENT_` in them at the start of newlines.

toolmantim:
Thanks @sj26 & @lox!

PR #44, opened 2018-06-07
Remove set -x as its too noisy, and breaks headers

toolmantim:
I spotted `set -x` in here. It might be a good idea to do without?


lox:
Yup yup, I've been meaning to remove this and make debug mode more useful.

toolmantim:
Thanks @lox!

PR #49, opened 2018-06-14
Fix bug with entrypoint

Pratima:
@toolmantim  Trying to fix the bug with entrypoint. Could you please validate?


toolmantim:
I'm trying to solve this with a new option in #50 — did you want to give that a test and see if it solves it for you? Warning: we might change it from `interpreter` to `shell` before it gets merged.

toolmantim:
Thanks for the PR! I ended up doing a bit more of an involved fix in #50 @Pratima, and you shouldn't need this anymore. I'll put out a release now.

toolmantim:
Okay, the fix should be out in [v1.3.0 (🍦 Gelato Hands)](https://github.com/buildkite-plugins/docker-buildkite-plugin/releases/tag/v1.3.0) @Pratima 🎉

Issue #23, opened 2018-04-23
Plugin assumes that bash is present in container


gh2k:
The command is run like this:
```
docker run "${args[@]}" "${BUILDKITE_PLUGIN_DOCKER_IMAGE}" bash -c "${BUILDKITE_COMMAND}"
```

When using Windows Server containers this doesn't work without some messing around. In my environment `bash -c` could be safely dropped. I would have expected that when adding the buildkite command to my pipeline file it would be the entire cmd section, rather than wrapped.

bash also isn't present in some minimal Linux containers IIRC (I think Alpine just has 'sh', or at least it used to)

Personally I'd prefer this to drop the `bash -c` entirely and assume that the pipeline author would add this if required. If there's a reason for this being there that I'm missing, perhaps you could provide a mechanism for overriding the interpreter in the config. For example:
```
  - command:  build-windows.ps1
    name: Windows build
    plugins:
      docker:
        image: microsoft/windowsservercore:1709
        always-pull: true
        interpreter: powershell -Command
    agents:
      windows: true
```

lox:
This is a really tricky one. See https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/124 for some more context.

toolmantim:
We could potentially also skip an interpreter for simple commands?

lox:
>We could potentially also skip an interpreter for simple commands?

@toolmantim yeah, that is a possibility. Perhaps we should just error for multi-line commands when there isn't a `shell` set and don't do it automatically for folks.

toolmantim:
Fixed in #50 @gh2k. I'll put out a release now. But we're going to do a major version bump to remove the default shell behaviour, and have the shell an opt-in thing, with some detections if people are trying to run bash-isms without having shell enabled.

PR #50, opened 2018-06-18
Add a shell option

toolmantim:
Similar to #49 this adds a `shell` option for disabling and customising the shell, and let's you run Docker commands in non-bash environments (#23).

It also disables the `shell` by default when you set an `entrypoint`, which is probably the most common need, which makes there no need for #49.


toolmantim:
Also I noticed we don't have `^` and `$` on many of our option pattern checks. Any reason for that @lox?

toolmantim:
@lox updated!

lox:
I'd make the default `/bin/bash -e -c` to match the agent.

toolmantim:
@lox this doesn't change the current behaviour at all though… I was thinking of doing that in a separate change. And just making it `/bin/sh -e -c`

lox:
Yeah, fair.

toolmantim:
This was released in [v1.3.0 (🍦 Gelato Hands)](https://github.com/buildkite-plugins/docker-buildkite-plugin/releases/tag/v1.3.0)

PR #66, opened 2018-09-07
Add support for Windows images

jeremiahsnapp:
This allows a user to use a Windows image such as `microsoft/windowsservercore` and set `windows-image: true` to ensure that proper Windows style paths and shell are used.

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -8,12 +8,26 @@ if [[ "${BUILDKITE_PLUGIN_DOCKER_DEBUG:-false}" =~ (true|on|1) ]] ; then
   debug_mode='on'
 fi
 
-args=(
-  "-it"
-  "--rm"
-  "--volume" "$PWD:${BUILDKITE_PLUGIN_DOCKER_WORKDIR:-/workdir}"
-  "--workdir" "${BUILDKITE_PLUGIN_DOCKER_WORKDIR:-/workdir}"
-)
+windows_image () { [[ "${BUILDKITE_PLUGIN_DOCKER_WINDOWS_IMAGE:-false}" =~ (true|on|1) ]]; }
+
+args=("--rm")
+
+if ! windows_image; then
+  args+=("-it")
+fi
+
+if windows_image; then
+  pwd=$(powershell -Command echo $PWD)
+  args+=(
+    "--volume" "$pwd:${BUILDKITE_PLUGIN_DOCKER_WORKDIR:-C:\workdir}"
+    "--workdir" "${BUILDKITE_PLUGIN_DOCKER_WORKDIR:-C:\workdir}"
+  )
+else
+  args+=(
+    "--volume" "$PWD:${BUILDKITE_PLUGIN_DOCKER_WORKDIR:-/workdir}"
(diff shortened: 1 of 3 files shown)


hooks/command
```diff
@@ -8,12 +8,26 @@ if [[ "${BUILDKITE_PLUGIN_DOCKER_DEBUG:-false}" =~ (true|on|1) ]] ; then
...
-  "--rm"
-  "--volume" "$PWD:${BUILDKITE_PLUGIN_DOCKER_WORKDIR:-/workdir}"
-  "--workdir" "${BUILDKITE_PLUGIN_DOCKER_WORKDIR:-/workdir}"
-)
+windows_image () { [[ "${BUILDKITE_PLUGIN_DOCKER_WINDOWS_IMAGE:-false}" =~ (true|on|1) ]]; }
+
+args=("--rm")
+
+if ! windows_image; then
+  args+=("-it")
```
jeremiahsnapp: `it` didn't work when the plugin was used on a Windows host. Honestly, I wonder if we can just get rid of `-it` altogether. How is it useful for linux build jobs?

lox:
I'm fine with removing `-it` generally, but I'm not sure about the addition of a `windows-image: true`. I could be talked into setting 'mount-buildkite-agent: false'  for windows by default though!

toolmantim:
I think without `-it` we don’t get lovely output for a bunch of commands like yarn etc. But I’m fully aware it’s just lies, because the process really shouldn’t think that it’s an interactive session.

jeremiahsnapp:
@lox In the interest of improving support for Windows builds I'm hoping we can find a solution that has good UX. I think having something like the `windows-image` property makes it much easier for people to use the plugin with Windows docker images. For example, it allows us to add `Set-Variable -Name ErrorActionPreference -Value Stop` before the build command by default so we get the same behavior we're used to from `bash -e`. It also automatically uses Windows path syntax `C:\...` for `$PWD` instead of the linux path syntax `/c/...` which wouldn't work at all.

If we can't have something like `windows-image` then this plugin would at least need some other way to modify the `$PWD` of the `--volume` flag so we can set it to the Windows path syntax. But requiring people to always have to set this manually doesn't feel like good UX to me.

If none of these options are desirable would you agree that there should be a separate plugin designed  specifically for Windows jobs? I was assuming you'd want to avoid that especially since most of this plugin works fine for Windows jobs as-is. It just needed these minor changes and a way to distinguish when to use them.

lox:
Sorry for the slow response @jeremiahsnapp. Have taken some time to think about this, and think I agree conceptually. We either need a flag for windows that changes several defaults of a different plugin. Let me think it over a bit more.

lox:
We're making some fairly major parameter changes presently, let's land those first.

lox:
@jeremiahsnapp we landed some of these things in https://github.com/buildkite-plugins/docker-buildkite-plugin/pull/76. We detect `tty` and `shell` for windows, as well as using correct paths and skipping the `buildkite-agent` mount. I'd love to hear your thoughts!

jeremiahsnapp:
Looks great @lox, thanks for the improvements! You might want to add a note in the README that `CMD` is the default shell for Windows. I wish the default shell for Windows was powershell instead but I'm sure you have good reasons for choosing it.

PR #67, opened 2018-09-14
Detect if a shell is needed for command

lox:
This defaults to no `shell`, but detects if a command is needed based on it containing shell-like chars and uses `/bin/sh -e -c`.

This uses the `xargs` shell-string parsing trick from https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/165. 

My only real concern here is that we match shell characters indiscriminately, so even something like `/my-command --flag="[blah]"` will trigger it, despite the square braces being in a quoted string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ Example: `docker`
 
 ### `network` (optional)
 
-Join the container to the docker network specified. The network will be created if it does not already exist. See https://docs.docker.com/engine/reference/run/#network-settings for more details. 
+Join the container to the docker network specified. The network will be created if it does not already exist. See https://docs.docker.com/engine/reference/run/#network-settings for more details.
 
 Example: `test-network`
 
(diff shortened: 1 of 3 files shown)


lox:
Yeah, the xargs stuff nicely tokenizes. I'm still uncomfortable that this will detect things like `/my/command "{"` as needing a shell though.

matthewd:
I think ```^([[:alnum:] =:./+_,\\-]|\\.|\'[^\']*\'|\"([^\\\"\$]|\\.)*\")*$``` might do it?

Or if you prefer to enumerate bad characters instead of good: ```^([^][\!\#\$\&\(\)\*\;\<\>\?\\\^\`\{\}\'\"]|\\.|\'[^\']*\'|\"([^\\\"\$]|\\.)*\")*$```... but I think the former is better. They also match bare newlines differently.

(I'll additionally note that the `xargs` trick doesn't like quoted strings that contain newlines, so we might want to treat those as shell-worthy too.)

lox:
After much thought, I reckon we are approaching this wrong. Closing this in favour of the approach in https://github.com/buildkite-plugins/docker-buildkite-plugin/pull/68.

PR #68, opened 2018-09-24
Add a command parameter and support arrays for command and shell

lox:
It dawned on @toolmantim and I that a good solution to the question of whether this plugin and `docker-compose` should default to a shell is the following:

* If a command is provided via the step-level command, assume a shell, as we tend to do things like `yarn install && yarn start` in examples:

```yml
steps:
  - command: yarn install && yarn run test
    plugins:
      docker#v1.4.0:
        image: "node:7"
        workdir: /app
```

The above will translate to `/bin/sh -e -c 'yarn install && yarn run test'` in the container. This will also support multiple line commands.

* If no step level command is provided, no shell is used. If you want control over exactly what command is passed to docker (e.g how the command is split into arguments), you can use the new plugin level `command` this PR introduces:

```yml
steps:
  - plugins:
      docker#v1.4.0:
        image: "koalaman/shellcheck"
        command: ["--exclude=SC2207", "/app/script.sh"]
        workdir: /app
```

This avoids us having to do any shell word parsing, which is horrible for complexity and security. 

The new array syntax is also applied to `entrypoint` and `shell`:

```yml
steps:
  - plugins:
       docker#v1.4.0:
         image: "alpine"
         command: ["echo", "hello", "world"]
         shell: ["/bin/sh", "-c"]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,17 @@ steps:
         workdir: /app
 ```
 
+If you want to control how your command is passed to the docker container, you can use the `command` parameter on the plugin directly:
+
+```yml
+steps:
+    plugins:
+      docker#v1.4.0:
+        image: "koalaman/shellcheck"
+        command: ["--exclude=SC2207", "/app/script.sh"]
+        workdir: /app
+```
+
 You can pass in additional environment variables:
 
 ```yml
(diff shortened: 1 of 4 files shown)


lox:
This is ready to go if you get a chance to review it @toolmantim or @matthewd

PR #69, opened 2018-10-01
Skip mounting the current checkout dir if custom mounts are provided

lox:
This changes the default behaviour of the plugin around how the current checkout directory is mounted into the container.

Currently, `$PWD` is mounted to whatever `workdir` is set to, which defaults to `/workdir`. With the changes in this PR, that step is skipped if custom volumes are provided. This let's power users choose what source directory is mounted in and where it goes in the container and leaves `workdir` to serve it's regular Docker purpose, which is to change the working directory in the docker container.

Additionally, this PR adds support for dot relative paths in the `mounts` mounts, so that things like `.:/app` work ok. 

**Breaking changes:** 
* Calls with custom `mounts` will now now not do the default "mount $PWD to workdir" behaviour

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,16 +8,18 @@ If you need more control, please see the [docker-compose Buildkite Plugin](https
 
 ## Example
 
-The following pipeline will run `yarn install` and `yarn run test` inside a Docker container using the [node:7 Docker image](https://hub.docker.com/_/node/):
+The following pipeline will build a binary in the dist directory using [golang:1.11 Docker image](https://hub.docker.com/_/golang/) and then uploaded as an artifact.
 
 ```yml
 steps:
-  - command: yarn install && yarn run test
+  - command: go build -o dist/my-app .
+    artifact_paths: ./dist/my-app
     plugins:
       docker#v1.4.0:
-        image: "node:7"
-        workdir: /app
+        image: "golang:1.11"
 ```
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -77,24 +72,24 @@ Example: `node:7`
...
-Whether to always pull the latest image before running the command. Useful if the image has a `latest` tag. The default is false, the image will only get pulled if not present.
-
 ### `mount-buildkite-agent` (optional)
 
 Whether to automatically mount the `buildkite-agent` binary from the host agent machine into the container. Defaults to `true`. Set to `false` if you want to disable, or if you already have your own binary in the image.
 
 ### `mounts` (optional)
 
-Extra volume mounts to pass to the docker container, in an array. Items are specified as `SOURCE:TARGET`. Each entry corresponds to a Docker CLI `--volume` parameter.
+Extra volume mounts to pass to the docker container, in an array. Items are specified as `SOURCE:TARGET`. Each entry corresponds to a Docker CLI `--volume` parameter, except relative paths are allowed. in `SOURCE`.
```
toolmantim: > in `SOURCE`

typo there?

toolmantim:
Related: I wonder if we should rename `mounts` to `volumes`? I'm guessing that's what people might be searching for if they've any experience with docker.

PR #70, opened 2018-10-01
Arrays only for command and shell

lox:
In the spirit of simplicity, this requires the array syntax to be provided for `command` and `shell`. Previous string values (which required tokenizing) will result in an error like:

> 🚨The Docker Plugin’s shell configuration option can no longer be specified as a string, but only as an array. Please update your pipeline.yml to use an array, for example: ["/bin/sh", "-e", "-u"].
>
> Note that the docker plugin will infer a shell if one is required, so you might be able to remove the option entirely.

**Breaking changes:**
* This will cause `exit 1` on most pipelines that use a `shell` directive.

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -2,24 +2,24 @@
 
 set -euo pipefail
 
-# Reads either a value or a list from plugin config into a global result array
+# Reads a list from plugin config into a global result array
 # Returns success if values were read
 function plugin_read_list_into_result() {
   local prefix="$1"
-  local parameter="${prefix}_0"
+  local i=0
+  local parameter="${prefix}_${i}"
   result=()
 
-  if [[ -n "${!parameter:-}" ]]; then
-    local i=0
-    local parameter="${prefix}_${i}"
-    while [[ -n "${!parameter:-}" ]]; do
-      result+=("${!parameter}")
-      i=$((i+1))
-      parameter="${prefix}_${i}"
-    done
-  elif [[ -n "${!prefix:-}" ]]; then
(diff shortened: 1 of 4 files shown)


hooks/command
```diff
@@ -2,25 +2,25 @@
...
-    local parameter="${prefix}_${i}"
-    while [[ -n "${!parameter:-}" ]]; do
-      result+=("${!parameter}")
-      i=$((i+1))
-      parameter="${prefix}_${i}"
-    done
-  elif [[ -n "${!prefix:-}" ]]; then
-    result+=("${!prefix}")
+  if [[ -n "${!prefix:-}" ]] ; then
+    echo "🚨 Plugin received a string for $prefix, expected an array" >&2
```
toolmantim: > 🚨 The Docker Plugin’s `$prefix` configuration option can no longer be specified as a string, but only as an array. Please update your pipeline.yml to use an array, for example: `["/bin/sh", "-e", "-u"]`.

It also had me wondering if we could somehow mention:

> The Docker Plugin now includes a shell by default, so you may be able to remove your `shell` configuration option altogether.

lox:
I've added some better error messages @toolmantim, we could possibly detect cases where a shell directive could be dropped. Not sure it would be worth it though 🤔

toolmantim:
@lox oh, I wasn't suggesting we even detect whether it could be dropped. But the transition point of string to array is a good chance for them to check whether they can drop it altogether… so a gentle nudge to check that might be good?

PR #72, opened 2018-10-02
Rename mounts to volumes

lox:
This renames `mounts` to `volumes` to align with docker terminology. 

It's backwards compatible, so the older `mounts` will still work.


lox:
@toolmantim possibly, I reckon we add a deprecation notice a bit later?

toolmantim:
👍🏼

PR #73, opened 2018-10-02
Allow `entrypoint:` to reset default entrypoint

### LINKED ISSUE #59: entrypoint should allow ""
https://github.com/buildkite-plugins/docker-buildkite-plugin/blob/master/hooks/command#L82 checks for a non-empty string in `entrypoint`, but it should allow `""` so you can reset the entrypoint without having to do weird things like `entrypoint: "/bin/sh"; shell: "-ec"`

lox:
Adds the ability to disable the Docker image’s entrypoint by setting `entry point: false`. This is equivalent to running `docker run --entrypoint ''`.

Closes #59.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,9 +137,9 @@ Enables debug mode, which outputs the full Docker commands that will be run on t
 
 Default: `false`
 
-### `entrypoint` (optional, string)
+### `entrypoint` (optional, string or boolean)
 
-Override the image’s default entrypoint, and defaults the `shell` option to `false`. See the [docker run --entrypoint documentation](https://docs.docker.com/engine/reference/run/#entrypoint-default-command-to-execute-at-runtime) for more details.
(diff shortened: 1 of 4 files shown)


lox:
I can't figure out a good way to workaround the parsing in the bats mock library to make this change work.

toolmantim:
Maybe we just have to comment/skip the test for this for now?

dgarbus:
Any update on this? We're running into a case where being able to set the entrypoint to nil would be helpful.

toolmantim:
I took a stab at this a few days ago, updated it to master, and had the same problem with `bats` and environment variables not passing through correctly, so I couldn't get the tests to work.

Though my version also supported `false` and `''` as valid values for disabling the entrypoint (by using `BUILDKITE_PLUGIN_DOCKER_PLUGIN_ENTRYPOINT:-undefined` as the default if check, instead of `false`).

Perhaps we just comment out the test, or find another way to test it?

lox:
I merged the updated plugin tester, lemme see if I can get those tests passing now.

pecigonzalo:
I was testing this PR and I believe its not working, I have tried setting `false` and `""` and it never sets `--entrypoint=""` which should do `--entrypoint ''`.
Could you show an example?

toolmantim:
It's possibly a bug @pecigonzalo.

PR #76, opened 2018-10-17
Disable tty and mount-buildkite-agent for macos/windows

lox:
This is one of the last breaking changes I want to make before 2.0.0. Currently if you run this plugin on windows or mac, you get errors by default due to the lack of tty on windows and the inability to mount `/usr/local/bin/buildkite-agent` with Docker for Mac or Docker for Windows. On windows, you also get errors because windows paths are needed for volume mounts. 

This adds a layer of operating system detection that can change the defaults for things. You can still opt to turn those features on manually, but out of the box you get defaults that work for mac/windows. 

Thoughts @toolmantim?

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Example: `/app`
 
 ### `mount-buildkite-agent` (optional)
 
-Whether to automatically mount the `buildkite-agent` binary from the host agent machine into the container. Defaults to `true`. Set to `false` if you want to disable, or if you already have your own binary in the image.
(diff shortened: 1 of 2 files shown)


lox:
@toolmantim Nah, that still won't work for the `buildkite-agent`. The binary would be in the wrong format. I considered trying to add a helpful message for the builds, but I think docker for mac does an ok job of it already.

toolmantim:
Ahh, yep. Makes sense!

lox:
I'm just going to do some windows testing.

lox:
I've changed quite a bit based on windows testing. Mind giving it another look over @toolmantim?

lox:
I'm going to merge this for now @toolmantim, but if you have any concerns, lemme know and we can address them in follow ups.

PR #80, opened 2018-10-24
Adds an option for propagating the environment to the container

### LINKED ISSUE #79: Add option for propagating all environment variables
Per the Elastic CI for AWS guidance (https://github.com/buildkite/elastic-ci-stack-for-aws#what-type-of-builds-does-this-support):

> Each project should encapsulate it's dependencies via Docker and Docker Compose

This results in using this plugin for setting up build dependencies. It would be nice to add an option to propagate *all* environment variables to the container, e.g.

```
steps:
  - command: "echo \$BUILDKITE_COMMIT"
    plugins:
      - docker#v2.0.0:
         image: bash 
```

Rather than having to specify that `BUILDKITE_COMMIT` propagates to the container (with `environment`)

zsims:
This closes #79 by adding a `propagate-environment` variable that exposes `$BUILDKITE_ENV_FILE` to the container (via `--envfile`)

The `$BUILDKITE_ENV_FILE` is the environment variables for the step, e.g. won't include `$PATH`, `$PWD` and other variables that may break/confuse the docker container.

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -158,6 +158,15 @@ while IFS='=' read -r name _ ; do
   fi
 done < <(env | sort)
 
+# Propagate all environment variables into the container if requested
+if [[ "${BUILDKITE_PLUGIN_DOCKER_PROPAGATE_ENVIRONMENT:-false}" =~ ^(true|on|1)$ ]] ; then
+  if [[ -n "${BUILDKITE_ENV_FILE:-}" ]] ; then
+    args+=( "--env-file" "${BUILDKITE_ENV_FILE:-}" )
+  else
+    echo -n "🚨 Not propagating environment variables to container as \$BUILDKITE_ENV_FILE is not set"
+  fi
+fi
+
 if [[ "${BUILDKITE_PLUGIN_DOCKER_ALWAYS_PULL:-false}" =~ ^(true|on|1)$ ]] ; then
   echo "--- :docker: Pulling ${BUILDKITE_PLUGIN_DOCKER_IMAGE}"
   docker pull "${BUILDKITE_PLUGIN_DOCKER_IMAGE}"
--- a/hooks/command
+++ b/hooks/command
@@ -161,7 +161,12 @@ done < <(env | sort)
 # Propagate all environment variables into the container if requested
 if [[ "${BUILDKITE_PLUGIN_DOCKER_PROPAGATE_ENVIRONMENT:-false}" =~ ^(true|on|1)$ ]] ; then
   if [[ -n "${BUILDKITE_ENV_FILE:-}" ]] ; then
-    args+=( "--env-file" "${BUILDKITE_ENV_FILE:-}" )
+    # Read in the env file and convert to --env params for docker
+    # This is because --env-file doesn't support newlines or quotes per https://docs.docker.com/compose/env-file/#syntax-rules
+    while read -r var; do
+      echo "${var%%=*}"
+      args+=( --env "${var%%=*}" )
+    done < "$BUILDKITE_ENV_FILE"
   else
     echo -n "🚨 Not propagating environment variables to container as \$BUILDKITE_ENV_FILE is not set"
   fi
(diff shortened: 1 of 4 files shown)


zsims:
Keen for ideas how this could also include variables exported in (earlier) plugin steps, and also agent environment variables. As `$BUILDKITE_ENV_FILE` only has the explicit step vars.

lox:
Sorry for the delay reviewing @zsims! I'm keen. One unfortunate thing is that the env file isn't compatible with docker's `env-file` directive. Docker's `env-file` parsed doesn't handle newlines in any way, shape or form, and it also won't strip quotes from variables. See https://github.com/buildkite/agent/pull/682 for some more context.

What needs to be done is using the env file to figure out what env vars should be set for the docker call like at https://github.com/buildkite/docker-bootstrap-example/blob/9270665850467f7614c291f1efd820e25dd443ea/bootstrap.sh#L36-L43.

lox:
I kind of liked `inherit-environment-vars` from https://github.com/chef/anka-buildkite-plugin. Thoughts @toolmantim?

lox:
Yup, you are right. `BUILDKITE_ENV_FILE` just contains env from the job sent to the agent: 

https://github.com/buildkite/agent/blob/6f3b7ceb80a6df0650052224eef70b61e599acb9/agent/job_runner.go#L233-L246

Figuring out what env had changed in previous hooks would be tricky, I think it would probably need some agent changes.

lox:
We could for instance also create a `BUILDKITE_HOOK_CHANGES_ENV_FILE` or similar.

lox:
Or we could figure out if it makes sense to add those changes to the existing `ENV_FILE`.

zsims:
Awesome, I'll look at updating this to parse the env file as you mentioned. Then look at the agent to expose `export`'d variables

zsims:
Updated to use `--env` instead of `--env-file` as you suggested, seems `env` and `$BUILDKITE_ENV_FILE` differ in format -- can see now why that's the case! 😄

zsims:
> We could for instance also create a BUILDKITE_HOOK_CHANGES_ENV_FILE or similar.
> Or we could figure out if it makes sense to add those changes to the existing ENV_FILE.

Any preference for the approach here? The buildkite agent assumes that `ENV_FILE` is immutable. Also would be nice to keep values out of the file (e.g. they might be secret) so I'm leaning towards `BUILDKITE_HOOK_CHANGES_ENV_FILE`

zsims:
I had a crack in https://github.com/buildkite/agent/pull/850

lox:
@toolmantim I'm thinking we merge this as-is?

zsims:
Sorry, that isn't true yet. I'll update the readme. It's just pipeline vars, and not those set in hooks

zsims:
Updated to clarify what gets propagated (e.g. what you see in the Buildkite UI). Should be good to land @lox 🕺

hooks/command
```diff
@@ -158,6 +158,20 @@ while IFS='=' read -r name _ ; do
   fi
 done < <(env | sort)
 
+# Propagate all environment variables into the container if requested
+if [[ "${BUILDKITE_PLUGIN_DOCKER_PROPAGATE_ENVIRONMENT:-false}" =~ ^(true|on|1)$ ]] ; then
+  if [[ -n "${BUILDKITE_ENV_FILE:-}" ]] ; then
+    # Read in the env file and convert to --env params for docker
+    # This is because --env-file doesn't support newlines or quotes per https://docs.docker.com/compose/env-file/#syntax-rules
+    while read -r var; do
```
lox: My last concern is that this will load in the value of env as it is at run time, vs what it was when things started. This might be confusing. 

For instance, if the ENV_FILE has llamas=always and a pre-command hook sets it to llamas=sometimes, I'd expect the initial value. 

Thoughts?

zsims: I'd expect the updated value, would make debugging a lot easier

lox:

PR #82, opened 2018-11-13
Don't convert paths on gitbash on windows

### LINKED ISSUE #81: Containers on windows don't seem to be running commands
Heya, I'm trying to setup a pipeline to run inside windows containers, and am getting some unexpected results. As far as I can tell, the commands aren't ran at all.

I'm running the build agent on a `Windows Server version 1803 Datacenter Core for Containers` image on Google Container Engine, on which I installed Git for Windows, NSSM and the build agent, as per https://buildkite.com/docs/agent/v3/windows.

My `pipeline.yml` looks like this
```
steps:    
  - label: windows-steps
    command: "echo hello"
    plugins:
      - docker#v2.0.0:
          image: "microsoft/dotnet:latest"
    agents:
      windows: true
```

And the output log looks like this:
```
Running plugin github.com/buildkite-plugins/docker-buildkite-plugin#v2.0.0 command hook
Running CMD.EXE /c 'echo hello' in microsoft/dotnet:latest | 2s
  | Microsoft Windows [Version 10.0.17134.345]
  | (c) 2018 Microsoft Corporation. All rights reserved.
  |  
  | C:\workdir>> cd C:\buildkite-agent\builds\gce-buildkite-agent-windows-1-1\angular\testsetup\c\buildkite-agent\builds\gce-buildkite-agent-windows-1-1\angular\testsetup
```

I thought the command might be running but the output not shown, so I tried something that should error out (`command: "not-a-binary"`) but the build still passed, same message, no error anywhere.

filipesilva:
Fix #81


lox:

PR #83, opened 2018-11-20
Allow running privileged containers

magec:
I need a way of creating privileged containers for running tests commands. This allows this.


lox:
Makes sense, thanks!

lox:

PR #89, opened 2019-01-04
Add mount-ssh-agent option

jam13:
I couldn't find any way of using git inside a docker container with the current plugin setup.

This adds a new option `mount-ssh-agent` (defaults to false) which mounts the ssh-agent in the container at `/ssh-agent` and sets the `SSH_AUTH_SOCK` environment variable. It also mounts `known_hosts`.

My initial attempt to do this using the stock plugin failed because `volumes` are only parsed at pipeline upload time and the ssh-agent socket changes from step to step. So this:
```
- label: 'Build'
  command:
    - ./build.sh
  plugins:
    docker#2.1.0:
      image: 'node:8'
      volumes:
      - .:/workdir
      - ${SSH_AUTH_SOCK}:/ssh-agent
      - ${HOME}/.ssh/known_hosts:/root/.ssh/known_hosts
      workdir: /workdir
      environment:
      - SSH_AUTH_SOCK=/ssh-agent
```
results in and invalid socket path being mounted.

Trying to work around this using pre-command hooks also proved fruitless.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,12 @@ Whether to automatically mount the `buildkite-agent` binary from the host agent 
 
 Default: `true` for Linux, and `false` for macOS and Windows.
 
+### `mount-ssh-agent` (optional, boolean)
+
+Whether to automatically mount the ssh-agent socket from the host agent machine into the container (at `/ssh-agent`and `/root/.ssh/known_hosts` respectively), allowing git operations to work correctly.
+
+Default: `false`
+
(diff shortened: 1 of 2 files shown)


lox:
This seems like a good idea to me, thanks for this @jam13! Could I trouble you to add a README entry for this?

jam13:
README updated.

lox:
Many thanks!

PR #90, opened 2019-01-05
Add ability to configure size of /dev/shm for running docker image

arifogel:
- add `shm-size` plugin setting
  - takes string which is then passed as `docker run --shm-size <string>`


toolmantim:
Thanks @arifogel! Can you add it to the readme too, and we’ll merge?

arifogel:
@toolmantim done.
FYI I wasn't sure how to run the test I added locally, so I'm still not sure I got the quoting right.

PR #92, opened 2019-01-22
Add support for `docker run --init`.

philwo:
This ensures that zombie processes are reaped inside the container.


lox:
Seems sensible to me, I wonder if we should consider a default of `false` initially? Otherwise it's probably a breaking change for anyone currently using containers with a `tini` endpoint

toolmantim:
If we reckon it's a new good default, we could potentially cut a new major version? May as well!

lox:
Ok cool!

philwo:
You're awesome. 😆 Thanks a lot for reviewing and merging!

PR #107, opened 2019-02-22
Add documentation for pull-retries

lox:


dhalperi:
Dumb Q: I looked at #95 and didn't see this keyword being linked to the environment variable read here: https://github.com/buildkite-plugins/docker-buildkite-plugin/pull/95/files#diff-7a7c160f9b73cfeaf4d8b93aee769fc1R228

Is the set of keywords that exist automatically turned into buildkite envs based on name? (If so, that's genius!)

lox:
@dhalperi yup, we expose config to plugins as env. `pull-retries` becomes `BUILDKITE_PLUGIN_DOCKER_PULL_RETRIES`. Check out https://buildkite.com/docs/pipelines/plugins#developing-a-plugin fora bit more context!

PR #114, opened 2019-03-28
Add propagate-uid-gid option

benesch:
Add an option, `propagate-uid-gid`, to match the container UID/GID to
the host UID/GID. This avoids littering shared mounts with root-owned
files. The existing `user` option isn't quite flexible enough, as it
requires you to know the name of the user you want to run as, and often
you just want to use whatever UID/GID the host agent happens to be
running as.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,12 @@ Whether or not to automatically propagate all pipeline environment variables int
 
 Note that only pipeline variables will automatically be propagated (what you see in the Buildkite UI). Variables set in proceeding hook scripts will not be propagated to the container.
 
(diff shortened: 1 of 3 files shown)


toolmantim:
Thanks @benesch! I just realised this doesn't have tests. Any chance you could contribute them? (https://github.com/buildkite-plugins/docker-buildkite-plugin#developing)

toolmantim:
Nevermind, I snuck one in via bacf92c so I can do a release :) Thanks so much for the PR!

benesch:
Woohoo, thanks @toolmantim!

PR #117, opened 2019-04-15
Beautify the "Running command [...]" message.

philwo:
The plug-in currently prints two messages that help with figuring out what it is actually running:
1) It always prints a "Running $command in $image" message.
2) In debug mode, it additionally prints the "docker run" command line, too.

There are two little problems with these messages:
1) If $command is very long, you get a very long group header (that also somehow sometimes doesn't quote the command correctly). It also gets messy when the command has multiple lines, because only the first line is printed as the group header, the rest then follows in the collapsed body.
2) The docker run command is printed without any context below the already printed command-line, which makes it a bit hard to understand what's going on.

The fixed version puts both messages in their own collapsed groups, which (IMHO) looks much nicer and also makes it easier to copy & paste them.

Before:
Screenshot 2019-04-15 at 17 08 12
Screenshot 2019-04-15 at 17 05 49

After:
Screenshot 2019-04-15 at 17 06 10
Screenshot 2019-04-15 at 17 06 21

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -352,11 +352,11 @@ elif [[ ${#command[@]} -gt 0 ]] ; then
   done
 fi
 
-echo "--- :docker: Running ${display_command[*]} in ${BUILDKITE_PLUGIN_DOCKER_IMAGE}"
+echo "--- :docker: Logging \"docker run\" command"
+echo "$ docker run ${args[*]}" >&2
 
-if [[ "${debug_mode:-off}" =~ (on) ]] ; then
-  echo "$ docker run ${args[*]}" >&2
-fi
+echo "--- :docker: Running command in ${BUILDKITE_PLUGIN_DOCKER_IMAGE}"
+echo "$ ${display_command[*]}" >&2
 
 # Don't convert paths on gitbash on windows, as that can mangle user paths and cmd options.
 # See https://github.com/buildkite-plugins/docker-buildkite-plugin/issues/81 for more information.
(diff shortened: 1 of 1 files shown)


lox:
Y'know, I was considering exactly this change the other day, been trying to figure out ways to reduce some of the logging output of builds.

I'm ok with this, @toolmantim any feels?

toolmantim:
This is awesome! 👏🏼 Pulling the command out of the header is a great idea. Thanks for the PR.

I was wondering if there was a case for always outputting the `docker run` command in the logs? It makes it easier to run something similar locally, and understand what the config is doing with the underlying commands.

Do you run with `debug: true` enabled most of the time @philwo?

philwo:
Indeed, we run with debug all the time, but just in order to get the docker run command for easier local reproduction 😅 I'd be in favor of just always printing it!

lox:
I'd be ok with always showing it too I think. Such a shame we have to enumerate all the env like that 😓

toolmantim:
Yeah, let’s always output it. Perhaps we don’t need a header?

What if we colorised the `docker run` command and args, so you could more easily scan the more interesting parts of it at a glance? Then perhaps it could go underneath the existing “Running command in …” header?

philwo:
@toolmantim Sounds like an interesting idea, but not sure how to do it...

Should we merge it like this for now and we could iterate on the idea with the colored output and see how that looks like?

@lox I've changed the PR to always print the message. :)

lox:
Yeah I reckon let's get this merged!

PR #121, opened 2019-04-29
Document mounting docker repo credentials

rcoup:
Otherwise if you want to push/pull images from remote docker repositories from within your step container it won't work.

For the Elastic Stack, the credentials on the host live at `/var/lib/buildkite-agent/.docker/config.json`

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ steps:
           volumes:
             - "/var/run/docker.sock:/var/run/docker.sock"
 ```
+* Note that to push/pull docker images within your step container, you'll additionally need to mount `~/.docker/config.json` – with the associated security implications.
 
 You can disable the default behaviour of mounting in the checkout to `workdir`:
 


toolmantim:
Thanks for taking the time to submit a PR.

Docker Credentials are stored in all sorts of ways these days, e.g. https://github.com/buildkite-plugins/docker-login-buildkite-plugin/pull/31, so I'm not sure if we can provide a good generic piece of advice/documentation on this sorry!

lox:
That was at my suggestion based on the current Elastic Stack @toolmantim, but I hear ya. Thanks @rcoup, we'll discuss.

rcoup:
@toolmantim so what's the approach with DOCKER_CONFIG ala the linked ticket? Is it as easy as:

```
volumes:
  - ${DOCKER_CONFIG}:/root/.docker/config.json
```
? or $$?

If that doesn't work could it the path(s) be changed to use a variable we _do_ have access to (ie. not the PID)?

toolmantim:
@rcoup sorry for the delayed reply! I don't think it'd be possible to use this plugin with `DOCKER_CONFIG` — neither `${DOCKER_CONFIG}` or `$${DOCKER_CONFIG}` would work correctly in the plugin config. We'd need to add an explicit parameter to mount it through for you perhaps?

I think we should close this PR for now @lox?

rcoup:
I'll close it, but it'd be nice to get some way to support people trying to make it work.

PR #122, opened 2019-04-30
Add docker sysctl support

moorthya:
Docker run has a sysctl option to specify kernel parameters in the container. This PR adds support for sysctl to the buildkite plugin. More information about this option can be found here: https://docs.docker.com/engine/reference/commandline/run/


lox:
Seems reasonable to me! Thanks @moorthya

PR #124, opened 2019-04-30
Update README to include sysctl

moorthya:
Forgot to update the README in my last PR when adding sysctl support.


toolmantim:
Oh sorry, we should have picked that up in review. Thanks for the PR! 🙌🏼

PR #125, opened 2019-05-02
Handle entrypoint for set and unset variables

pecigonzalo:
Fixes: #73

PR #131, opened 2019-06-04
Add entrypoint disabling tests from #125 (thanks @pecigonzalo)

lox:
This adds the tests for disabling entrypoint with `entrypoint: ''` or `entrypoint: false`.

### DIFF
--- a/tests/command.bats
+++ b/tests/command.bats
@@ -331,44 +331,35 @@ EOF
   unstub docker
 }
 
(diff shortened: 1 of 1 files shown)


lox:
Am I missing something @pecigonzalo? Shouldn't these be failing?

pecigonzalo:
I replied in #125 but yes, they have false positives, because BATS is doing some funny stuff apparently.

Issue #105, opened 2019-02-18
Add `--add-host` functionality


jahredhope:
## Feature request
Allow the consumer to pass in values to be applied with `--add-host="host:IP` to append records to `/etc/hosts`.

As described in [Docker Network settings](https://docs.docker.com/engine/reference/run/#/network-settings)

#### Suggested syntax
```yaml
  docker#v1.4.0:
    image: 'node:10.15'
    hosts:
     - 'example.com.local:127.0.0.1'
# OR
       example.com.local: '127.0.0.1'
```

PR #133, opened 2019-06-06
Add support for the `add-host` flag on `docker run`

underscorebrody:
Resolves https://github.com/buildkite-plugins/docker-buildkite-plugin/issues/105

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -212,6 +212,13 @@ while IFS='=' read -r name _ ; do
   fi
 done < <(env | sort)
 
+# Parse host mappings and add them to the docker args
+while IFS='=' read -r name _ ; do
+  if [[ $name =~ ^(BUILDKITE_PLUGIN_DOCKER_ADD_HOST_[0-9]+) ]] ; then
+    args+=( "--add-host" "${!name}" )
+  fi
+done < <(env | sort)
+
 # Privileged container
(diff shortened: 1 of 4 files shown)


hooks/command
```diff
@@ -212,6 +212,13 @@ while IFS='=' read -r name _ ; do
   fi
 done < <(env | sort)
 
+# Parse host mappings and add them to the docker args
+while IFS='=' read -r name _ ; do
```
toolmantim: This isn't really a question related to this PR, but I wanted to ask @lox: is there any reason we prefer this syntax, over a simpler for loop?

```shell
i=0
while true; do
  env_name="BUILDKITE_PLUGIN_DOCKER_ADD_HOST_$i"
  if [[ -z "${!env_name:-}" ]]; then break; fi
  args+=( "--add-host" "${!env_name}" )
  i=$((i+1))
done
```

lox: I use this pattern when I want to refer to the $i later on.

lox: Oh, sorry, I see what you mean.

lox: I tend to use the while read loop for most things because I understand it's edge cases really well and usually I am iterating over output from another command and this is the only way to set arguments in the scope outside of the while loop.

In this example, I'd be totally 👌🏻 for a simpler loop.

lox:
Looks legit to me! Sorry for the delay.

hooks/command
toolmantim: Cool! That makes total sense @lox!

PR #135, opened 2019-07-04
Add support for --device flags on docker run

omeid:
This pull requests adds support for `--device` flag for docker run.

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -135,6 +135,19 @@ if plugin_read_list_into_result BUILDKITE_PLUGIN_DOCKER_VOLUMES BUILDKITE_PLUGIN
   done
 fi
 
+# Ensure devices option is an array
+if [[ -n "${BUILDKITE_PLUGIN_DOCKER_DEVICES:-}" ]] ; then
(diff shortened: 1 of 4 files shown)


hooks/command
```diff
@@ -135,6 +135,19 @@ if plugin_read_list_into_result BUILDKITE_PLUGIN_DOCKER_VOLUMES BUILDKITE_PLUGIN
   done
 fi
 
+# Ensure devices option is an array
```
toolmantim: I don't think we need these array checks anymore, because `plugin_read_list_into_result` already checks it and outputs an error.

omeid: Thanks for the swift feedback. I am happy to remove this instance of it but perhaps a more through cleanup is a better approach?

toolmantim: Sure thing! Let's just go with it.

omeid: Cool, can you please confirm that accepting `boolean` for `volumes` is a mistake? so that I can safely remove that check too.

omeid:
@toolmantim Thanks for merging this in, I have created #136 as a follow up on the cleanup.

Cheers

PR #139, opened 2019-07-30
Allow for agent specific vars in volumes and env vars

srmocher:
We would like to leverage agent specific metadata described by `BUILDKITE_AGENT_META_DATA_*` in mounting agent specific volumes and passing in agent-specific env var values when spinning up a docker container. 

e.g
```
volumes:
   - "/path/to/$$BUILDKITE_AGENT_NAME:/path/inside/container"
```
The value of `BUILDKITE_AGENT_NAME` doesn't get evaluated by the docker plugin currently at runtime. This change allows us for interpolation of variables in working directory, volumes and env passed to the plugin.

Not a bash expert - so maybe there's a better way to expand/interpolate the param values.


lox:
We hear you on this one @srmocher, it's something we want to add more holistically with support for a new kind of variables that are interpolated at run time.

In the short term, I'm not keen on solving the problem with an eval, it leads to all sorts of follow on issues and potential security risks.

PR #141, opened 2019-09-25
Support --userns

elruwen:
User namespaces (https://docs.docker.com/engine/security/userns-remap/) are a useful security tool. Sometimes you need fine grain control over them and you want to choose a different one than the default one. This change enables that.


lox:
Looks super legit! Sorry for the slow reply!

PR #143, opened 2019-11-01
Add some better errors when $SSH_AUTH_SOCK is wrong

lox:
Currently we don't provide any helpful errors when users use `mount-ssh-agent` and the `$SSH_AUTH_SOCK` either doesn't exist or points to something that isn't readable.

This adds some better errors.

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -172,6 +172,14 @@ fi
 
 # Mount ssh-agent socket and known_hosts
 if [[ "${BUILDKITE_PLUGIN_DOCKER_MOUNT_SSH_AGENT:-$mount_ssh_agent}" =~ ^(true|on|1)$ ]] ; then
+  if [[ -z "${SSH_AUTH_SOCK:-}" ]] ; then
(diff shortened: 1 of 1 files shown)


hooks/command
```diff
@@ -172,6 +172,14 @@ fi
 
 # Mount ssh-agent socket and known_hosts
 if [[ "${BUILDKITE_PLUGIN_DOCKER_MOUNT_SSH_AGENT:-$mount_ssh_agent}" =~ ^(true|on|1)$ ]] ; then
+  if [[ -z "${SSH_AUTH_SOCK:-}" ]] ; then
+    echo "+++ 🚨 \$SSH_AUTH_SOCK isn't set, has ssh-agent started?"
+    exit 1
+  fi
+  if [[ ! -f "${SSH_AUTH_SOCK}" ]] ; then
```
plasticine: Can this is `-S` to test for socket? Or perhaps even a different test after the file check that specifically tests the socket-ness of the file?

lox: Yeah, good idea @plasticine

PR #152, opened 2020-01-07
Escape /C to workaround a bug in Git for Windows 2.24.1.2

jeremiahsnapp:
This change works with Git for Windows 2.24.1.2 and backwards
compatibility test with 2.22.0 was also successful.

https://github.com/git-for-windows/git/issues/2442
Signed-off-by: Jeremiah Snapp <jeremiah@chef.io>


toolmantim:
Thanks @jeremiahsnapp!

PR #153, opened 2020-01-23
Make command syntax consistent across different examples

Zhong-z:
Fix typo


toolmantim:
Thanks @Zhong-z! On second review, this consistency fix looks fine… both `command` and `commands` are the same at that level of the pipeline.

@pda the difference you're referring to is the plugin config option, vs the step-level config key — confusing I know!

pda:
Ah yes 🤦‍♂️

PR #156, opened 2020-04-17
Add clarification to docs around S3 secrets plugin interaction

dannykansas:
After a great bit of trial and error, I discovered that the `propagate-environment` flag does not, in fact, result in proper loading of secrets injected by [Buildkite's S3 Secrets plugin](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks).

This PR adds adds clarification to documentation explaining how:

- Environment variables imported from the [Buildkite S3 secrets plugin](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks) will not properly load unless explicitly defined
- The `propagate-environment` flag does not automatically import environment variables defined in the [S3 secrets plugin](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks) environment hook

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,9 @@ steps:
     artifact_paths: "dist/**"
 ```
 
-You can pass in additional environment variables and customize what is mounted into the container:
+You can pass in additional environment variables and customize what is mounted into the container. 
+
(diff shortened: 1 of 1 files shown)


nithyaasworld:
Yeah, propagate-environment flag would only propagate environment variables available in the step as stated in the docs. To import other environment variables , we need to explicitly pass in its key names under the 'environment' attribute of the docker plugin. 

It's always good to have more clarification. Thanks @dannykansas for adding more clarification!

nithyaasworld (approved):

PR #159, opened 2020-05-13
Allow host networking

petemounce:
https://docs.docker.com/network/network-tutorial-host/#procedure

`host` is a special network and doesn't need creating, I think?


jeremybumsted:
Apologies for getting back to you so late on this PR!

@petemounce you are correct, the `host` network exists by default so it doesn't need to be created. Because the plugin checks for existing networks already, it already allows the host network to be used, it just returns a message saying `docker network host already exists`

PR #162, opened 2020-06-19
Add cpus argument

dieend:
Add cpus argument on docker run:
https://docs.docker.com/config/containers/resource_constraints/#configure-the-default-cfs-scheduler


toolmantim (approved):
Thanks so much @dieend, and sorry for the delay! I've just added a short readme for the new option, but otherwise it looks great. I'll merge and release now.

PR #167, opened 2020-09-28
Support buildkite-agent's git-mirrors; mount BUILDKITE_GIT_MIRROR_PATH.

pda:
When `BUILDKITE_GIT_MIRROR_PATH` environment is present, that path is automatically mounted to the same location within the container. This makes repos created with `git clone --reference $BUILDKITE_GIT_MIRROR_PATH` accessible within the docker container.

Until https://github.com/buildkite/agent/pull/1311 the agent does not set `BUILDKITE_GIT_MIRROR_PATH`, ~so this pull request is a draft / RFC~.
This functionality will depend on (upcoming) buildkite-agent >= [v3.24.0](https://github.com/buildkite/agent/releases/tag/v3.24.0).

_Same as buildkite-plugins/docker-compose-buildkite-plugin#287 but this one is for docker not docker-compose._

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ steps:
   - command: "go build -o dist/my-app ."
     artifact_paths: "./dist/my-app"
     plugins:
-      - docker#v3.5.0:
+      - docker#v3.7.0:
           image: "golang:1.11"
 ```
 
@@ -23,7 +23,7 @@ Windows images are also supported:
 steps:
   - command: "dotnet publish -c Release -o published"
     plugins:
-      - docker#v3.5.0:
+      - docker#v3.7.0:
           image: "microsoft/dotnet:latest"
           always-pull: true
 ```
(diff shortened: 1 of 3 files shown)


ngan:
Nice! This plugin also has `propagate-environment` option. Wondering if this should pass env var as well.

pda:
Hmm.
I would lean towards not including the new `BUILDKITE_REPO_MIRROR` in `propagate-environment` because:
1. The documentation says “Note that only pipeline variables will automatically be propagated (what you see in the Buildkite UI)” an this new var doesn't fit that description, and
2. I can't think of a use-case for `BUILDKITE_REPO_MIRROR` inside the container. The git repo will already have it in `.git/objects/info/alternates`, so you can always grab it from there if you need it?

ngan:
That’s fair. 👍

pda:
(@ngan I wrote that in a hurry, and intended to ask if you had a use-case for `BUILDKITE_REPO_MIRROR` inside the container? Or just future proofing in case you need it?)

ngan:
Nah, like you I don’t see any reason either. Just wanted to bring it up in case you might have overlooked it. All good on my end.

PR #172, opened 2020-11-30
Added gpus option

a-canela:
The GPUs Docker option was added to this Buildkite plugin. More details about this Docker option can be found here:

https://docs.docker.com/config/containers/resource_constraints/#gpu


pzeballos:
Thanks @a-canela! I did a couple of extra commits to fix the build 🙂 Sorry for the delay! 🙏🏻

PR #173, opened 2021-01-11
Add a container name

albertyw:
Partially copied from the [docker-compose-buildkite-plugin](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/master/commands/run.sh#L84-L85), this adds a useful container name rather than allowing the docker daemon to choose one randomly.

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -343,6 +343,10 @@ elif plugin_read_list_into_result BUILDKITE_PLUGIN_DOCKER_SHELL ; then
   done
 fi
 
+# Add a container name
+container_name="${BUILDKITE_PIPELINE_SLUG}_${BUILDKITE_BUILD_NUMBER}"
+args+=("--name" "$container_name")
+
 # Add the job id as meta-data for reference in pre-exit
 args+=("--label" "com.buildkite.job-id=${BUILDKITE_JOB_ID}")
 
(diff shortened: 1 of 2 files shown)


hooks/command
```diff
@@ -343,6 +343,10 @@ elif plugin_read_list_into_result BUILDKITE_PLUGIN_DOCKER_SHELL ; then
   done
 fi
 
+# Add a container name
+container_name="${BUILDKITE_PIPELINE_SLUG}_${BUILDKITE_BUILD_NUMBER}"
```
pda, line 347: The docker-compose plugin integrates `BUILDKITE_JOB_ID` into the container name, which means the name is unique per job (which includes being different across retries of a failed build step etc):

https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/2f44a9c309336df0558d21b67770f5f89f4002fa/lib/shared.bash#L84

I think we might need to do something similar here?
Otherwise a single agent machine (with a single docker daemon) could concurrently run two jobs in the same build that end up with the same `container_name`.

pda (changes requested):
Good idea — I've left a note about keeping the names unique in different jobs.

nsuma8989:
Hi @albertyw Please let us know if you are looking into the changes suggested in the review and update the pull request ?

pzeballos:
Hey! Feel free to open a new one with the changes proposed. Thanks!

PR #177, opened 2021-01-11
Add "copy-checkout" config

### LINKED ISSUE #113: copy-checkout
We've run into situations where the directory mounted in the container gets into a bad state, and the buildkite-agent user isn't privileged enough to delete the contents of the directory.

We had a couple solutions in mind, one of the most reasonable of which we thought would be using the `docker cp` command to put the checked-out directory into the container. Another way we could accomplish this is by injecting a few commands to mount read-only volumes and copying them through the container to avoid any changes propogating, but `docker cp` seems like a cleaner solution.

Would be willing to put in a PR to mirror the usage of `mount-checkout`, but for `docker cp`'ing a directory into the container/image at runtime. Would like to hear your thoughts!

albertyw:
This adds a `copy-checkout` option that is similar to the `mount-checkout` option.  Instead of mounting the working directory of test code, it copies the code into the container so that I can modify the test directory without polluting the host's copy.  

I found that I needed this because I was running into one of a few scenarios when trying to install test dependencies:

1.  With default options, I would write dependencies into the container but it would break the next test because of root-owned files that the buildkite-agent wouldn't be able to delete.  
2. I tried running with `propagate-uid-gid` but then the user within the container would not have root permissions, needed for things like `apt-get install`.
3.  I could run with `mount-checkout:false` to not pollute the host's copy of the working directory, but I'd have to go through the kerfluffle of installing git, cloning the correct repository, checking out the correct branch, etc.
4. I could also manually run post-test cleanup or immediately copy the working directory at the beginning of the test, but that seems hacky.  

Therefore, I think the cleanest way for me is to use this `copy-checkout` config to get a copy of the code and have it isolated from the host's copy.  

(I have yet to test out the docker-compose-buildkite-plugin, but maybe it also provides a workaround to my problem)

Fixes #113
Depends on #173

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -18,5 +18,5 @@ steps:
   - label: run the plugin
     command: "echo hello world"
     plugins:
-      ${BUILDKITE_REPO}#${BUILDKITE_COMMIT}:
+      ${BUILDKITE_PULL_REQUEST_REPO:-$BUILDKITE_REPO}#${BUILDKITE_COMMIT}:
         image: ubuntu:18.04
--- a/README.md
+++ b/README.md
@@ -185,6 +185,12 @@ Whether to automatically mount the current working directory which contains your
 
 Default: `true`
 
+### `copy-checkout` (optional, boolean)
+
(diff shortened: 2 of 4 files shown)


albertyw:
It looks like this currently isn't going to work as intended because the `docker run` runs the test before the `docker copy` runs.

PR #181, opened 2021-01-22
Add storage opts

bennetty:
add storage opts flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,7 +233,7 @@ Example: `2gb`
 
 ### `storage-opt` (optional, string)
 
-This allows to set the container rootfs size at the creation time. This is only available for the devicemapper, btrfs, overlay2, windowsfilter and zfs graph drivers. See [docker documentation](https://docs.docker.com/engine/reference/commandline/run/#set-storage-driver-options-per-container) for more details.
+This allows setting the container rootfs size at the creation time. This is only available for the `devicemapper`, `btrfs`, `overlay2`, `windowsfilter` and `zfs` graph drivers. See [docker documentation](https://docs.docker.com/engine/reference/commandline/run/#set-storage-driver-options-per-container) for more details.
 
 Example: `size=120G`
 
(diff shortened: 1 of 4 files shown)


nithyaasworld:
Thanks @bennetty for the PR! Sorry that it took us a long time to get to this. We verified it and we would be merging this soon. 

Could you also please update the Readme file with an example for this? Thanks again!

nithyaasworld:
While the plugin works, the [storage-opts flag](https://docs.docker.com/engine/reference/commandline/run/#set-storage-driver-options-per-container) is added as an array though it supposed to have only one value. Closing this PR. @bennetty  If you would like, please feel free to re-open this PR with updated changes that would accept storage-opts flag as string rather than an array.

plugin.yml
```diff
@@ -65,6 +65,8 @@ configuration:
       type: boolean
     sysctls:
       type: array
+    storage-opt:
+      type: string
```
pzeballos, line 69: Makes sense! 👌
We probably should order attributes in alphabetical order ... but that's outside this PR

README.md
```diff
@@ -231,6 +231,12 @@ Set the size of the `/dev/shm` shared memory filesystem mount inside the docker
 
 Example: `2gb`
 
+### `storage-opt` (optional, string)
+
+This allows to set the container rootfs size at the creation time. This is only available for the devicemapper, btrfs, overlay2, windowsfilter and zfs graph drivers. See [docker documentation](https://docs.docker.com/engine/reference/commandline/run/#set-storage-driver-options-per-container) for more details.
```
pzeballos, line 236: I would add a backtick on each driver name:

```suggestion
This allows setting the container rootfs size at the creation time. This is only available for the `devicemapper`, `btrfs`, `overlay2`, `windowsfilter` and `zfs` graph drivers. See [docker documentation](https://docs.docker.com/engine/reference/commandline/run/#set-storage-driver-options-per-container) for more details.
```

pzeballos (changes requested):
Only a small change on the README, but the rest LGTM

pzeballos (approved):

pzeballos:
Thanks @bennetty for your contributions and sorry for the delay!!

PR #182, opened 2021-01-23
Add "copy-checkout" config

### LINKED ISSUE #113: copy-checkout

albertyw:
This adds a `copy-checkout` option that is similar to the `mount-checkout` option.  Instead of mounting the working directory of test code, it copies the code into the container so that I can modify the test directory without polluting the host's copy.  

I found that I needed this because I was running into one of a few scenarios when trying to install test dependencies:

1.  With default options, I would write dependencies into the container but it would break the next test because of root-owned files that the buildkite-agent wouldn't be able to delete.  
2. I tried running with `propagate-uid-gid` but then the user within the container would not have root permissions, needed for things like `apt-get install`.
3.  I could run with `mount-checkout:false` to not pollute the host's copy of the working directory, but I'd have to go through the kerfluffle of installing git, cloning the correct repository, checking out the correct branch, etc.
4. I could also manually run post-test cleanup or immediately copy the working directory at the beginning of the test, but that seems hacky.  

Therefore, I think the cleanest way for me is to use this `copy-checkout` config to get a copy of the code and have it isolated from the host's copy.  

(I have yet to test out the docker-compose-buildkite-plugin, but maybe it also provides a workaround to my problem)

Fixes #113
Depends on #173 
Replaces #177 which was an approach that doesn't work.

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -130,6 +130,11 @@ if [[ "${BUILDKITE_PLUGIN_DOCKER_MOUNT_CHECKOUT:-on}" =~ ^(true|on|1)$ ]] ; then
   args+=( "--volume" "${pwd_default}:${workdir}" )
 fi
 
+# Mount $PWD onto a new WORKDIR if mount-checkout is true
+if [[ "${BUILDKITE_PLUGIN_DOCKER_COPY_CHECKOUT:-off}" =~ ^(true|on|1)$ ]] ; then
+  args+=( "--volume" "${pwd_default}:${workdir}_original" )
+fi
+
 # Parse volumes (and deprecated mounts) and add them to the docker args
 if plugin_read_list_into_result BUILDKITE_PLUGIN_DOCKER_VOLUMES BUILDKITE_PLUGIN_DOCKER_MOUNTS ; then
   for arg in "${result[@]}" ; do
@@ -362,6 +367,11 @@ args+=("--label" "com.buildkite.job-id=${BUILDKITE_JOB_ID}")
 
 # Add the image in before the shell and command
 args+=("${BUILDKITE_PLUGIN_DOCKER_IMAGE}")
+
+# Set up the copied workdir if copy-checkout
+if [[ "${BUILDKITE_PLUGIN_DOCKER_COPY_CHECKOUT:-off}" =~ ^(true|on|1)$ ]] ; then
+  args+=("cp" "-r" "${pwd_default}" "${workdir}_original" "&&" )
+fi
 
 # Set a default shell if one is needed
 if [[ -z $shell_disabled ]] && [[ ${#shell[@]} -eq 0 ]] ; then
--- a/hooks/command
+++ b/hooks/command
@@ -368,11 +368,6 @@ args+=("--label" "com.buildkite.job-id=${BUILDKITE_JOB_ID}")
 # Add the image in before the shell and command
 args+=("${BUILDKITE_PLUGIN_DOCKER_IMAGE}")
 
-# Set up the copied workdir if copy-checkout
-if [[ "${BUILDKITE_PLUGIN_DOCKER_COPY_CHECKOUT:-off}" =~ ^(true|on|1)$ ]] ; then
-  args+=("cp" "-r" "${pwd_default}" "${workdir}_original" "&&" )
-fi
-
 # Set a default shell if one is needed
 if [[ -z $shell_disabled ]] && [[ ${#shell[@]} -eq 0 ]] ; then
   if is_windows ; then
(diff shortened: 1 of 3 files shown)


hooks/command
```diff
@@ -360,6 +379,17 @@ fi
 
 command=()
 
+# Set up the copied workdir if copy-checkout
+if [[ "${BUILDKITE_PLUGIN_DOCKER_COPY_CHECKOUT:-off}" =~ ^(true|on|1)$ ]] ; then
+  COPY_COMMAND="cp -r ${workdir}_original/* ${workdir}"
+  if [[ -n "${BUILDKITE_COMMAND}" ]] ; then
+    BUILDKITE_COMMAND="${COPY_COMMAND} && ${BUILDKITE_COMMAND}"
```
albertyw, line 386: This should work correctly

hooks/command
```diff
@@ -360,6 +379,17 @@ fi
...
+  fi
+  if [[ -n "${BUILDKITE_PLUGIN_DOCKER_COMMAND_0:-}" ]] ; then
+    command+=("$COPY_COMMAND && ")
```
albertyw, line 389: I haven't been able to get prepending the copy command to work correctly.  The `&&` keeps getting escaped and I'm not good enough at bash to correctly unescape them so it can execute both the copy and the user-supplied commands.  I could use some pointers here.

hooks/command
```diff
@@ -111,15 +111,30 @@ fi
...
+fi
+
 # By default, mount $PWD onto $WORKDIR
 if [[ "${BUILDKITE_PLUGIN_DOCKER_MOUNT_CHECKOUT:-on}" =~ ^(true|on|1)$ ]] ; then
   args+=( "--volume" "${pwd_default}:${workdir}" )
 fi
 
+# Mount $PWD onto a new WORKDIR if mount-checkout is true
+if [[ "${BUILDKITE_PLUGIN_DOCKER_COPY_CHECKOUT:-off}" =~ ^(true|on|1)$ ]] ; then
+  args+=( "--volume" "${pwd_default}:${workdir}_original" )
```
pda, line 135: Should this be mounted read-only with a `:ro` suffix?

hooks/command
```diff
@@ -360,6 +379,17 @@ fi
 
 command=()
 
+# Set up the copied workdir if copy-checkout
+if [[ "${BUILDKITE_PLUGIN_DOCKER_COPY_CHECKOUT:-off}" =~ ^(true|on|1)$ ]] ; then
+  COPY_COMMAND="cp -r ${workdir}_original/* ${workdir}"
```
pda, line 384: I don't think `${workdir}_original/*` will match hidden (dot) files/directories?

pda: I don't think `cp -r` preserves various attributes that might be important… you might want `cp -a`?

Also, could `$workdir` have spaces in it? In which case you might have trouble with it being interpreted as multiple separate arguments to `cp`. It sure gets tricky trying to glue together strings as commands in bash 😬

pda (commented):
I think I like the intention here.
I think there might be some bash gotchas though.

pda:
I wonder if there's a simpler way to copy the workdir into the container before/during startup, rather than as a volume / command injection.

#113 mentions using `docker cp`, but I gather that would need to be part of a multi-phase process of `docker create …` → `docker cp …` → `docker start`, instead of a simple `docker run …`. But maybe it's viable.

Or, do a `docker build` with Dockerfile from stdin, creating a temporary image to run instead of the upstream one:

```
$ echo -e "FROM alpine\nCOPY . /workdir/" | docker build --tag tmp_image --file - .
Sending build context to Docker daemon  3.119kB
Step 1/2 : FROM alpine
 ---> a24bb4013296
Step 2/2 : COPY . /workdir/
 ---> Using cache
 ---> a2467f75f2d4
Successfully built a2467f75f2d4
Successfully tagged tmp_image:latest

$ docker run --rm tmp_image ls -al /workdir
total 12
drwxr-xr-x    2 root     root          4096 Mar  2 06:13 .
drwxr-xr-x    1 root     root          4096 Mar  2 06:13 ..
-rw-r--r--    1 root     root             6 Mar  2 06:13 hello
-rw-r--r--    1 root     root             0 Mar  2 06:13 world
```

pzeballos:
Hi @albertyw! Are you up on finishing the PR? Thanks!

pzeballos:
If you are still interested in this functionality, you can create a new PR with the changes suggested, and we'll review it again 😊 Thanks!

PR #183, opened 2021-02-16
Adds a setting to allow propagating AWS auth tokens (from an assume role plugin) to the container.

chrisfowles:
This adds a setting `propagate-aws-auth-tokens` which will propagate the following environment variables to the container if they are set:
`AWS_ACCESS_KEY_ID`
`AWS_SECRET_ACCESS_KEY`
`AWS_SESSION_TOKEN`

This allows you to use assume role plugins (which don't expose these via the `BUILDKITE_ENV_FILE` without having to specify `environment` all the time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,12 @@ Whether or not to automatically propagate all pipeline environment variables int
 
 Note that only pipeline variables will automatically be propagated (what you see in the Buildkite UI). Variables set in proceeding hook scripts will not be propagated to the container.
 
+### `propagate-aws-auth-tokens` (optional, boolean)
+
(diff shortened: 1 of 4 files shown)


pda (approved):
Great idea, and beautiful implementation, docs & tests 👌🏼

Could perhaps argue for setting `AWS_ACCESS_KEY_ID` etc in one of the tests that doesn't set `propagate-aws-auth-tokens: true`, to assert that they're not leaking through regardless of the new option. But the code path is nice and simple, I don't think that's a likely future regression. 🚀

jayco (approved):
Looks great - thanks @chrisfowles 🎉

Issue #138, opened 2019-07-24
entrypoint: false doesn't disable default image entrypoint


orien:
In v3.3.0, the [`entrypoint` documentation](https://github.com/buildkite-plugins/docker-buildkite-plugin/blob/v3.3.0/README.md#entrypoint-optional-string-or-boolean) specifies:

> Set it to `false` to disable the default entrypoint for the image.

However, doing so doesn't disable the entrypoint and I see no `--entrypoint` flag in the logged `docker run` command.

```yaml
steps:
  - command: ls
    plugins:
      - docker#v3.3.0:
          image: hashicorp/terraform:light
          entrypoint: false
```
```
docker run -it --rm --init --volume /var/lib/buildkite-agent/builds/xxxx:/workdir --workdir /workdir --env BUILDKITE_JOB_ID --env BUILDKITE_BUILD_ID --env BUILDKITE_AGENT_ACCESS_TOKEN --volume /usr/bin/buildkite-agent:/usr/bin/buildkite-agent --label com.buildkite.job-id=6b701e83-acb8-4834-a129-ccb3266577a7 hashicorp/terraform:light /bin/sh -e -c ls
```

Here the default entrypoint is being used and for this image ends up running `terraform /bin/sh -e -c ls`. This fails because `/bin/sh` is an invalid subcommand for `terraform`.

#### Work-around

Setting this attribute to an empty string seems to produce the desired behaviour.

```yaml
entrypoint: ""
```

```
docker run -it --rm --init --volume /var/lib/buildkite-agent/builds/xxxx:/workdir --workdir /workdir --env BUILDKITE_JOB_ID --env BUILDKITE_BUILD_ID --env BUILDKITE_AGENT_ACCESS_TOKEN --volume /usr/bin/buildkite-agent:/usr/bin/buildkite-agent --entrypoint '' --label com.buildkite.job-id=107582c8-92e5-4a95-8e34-93523f5c6a82 hashicorp/terraform:light ls
```

#### Suggestion

I think we should update the plugin behaviour to match the documentation (alternatively, update the documentation to match the plugin behaviour).

toolmantim:
Sorry you've hit this bug. I believe this is being addressed in #132, though it hasn't been updated in while.

kylecordes:
Ouch, this one was tricky. After some failed experimentation this bug report was vital reading.  I suggest a quick fix to the docs until there is time to fix the problem.

[...]

PR #132, opened 2019-06-05
Set an empty string for entrypoint if false-y

lox:
Another take on #125.

PR #188, opened 2021-09-21
Update README with workaround for entrypoint option

bunsenmcdubbs:
Until #138 is resolved (possibly by #132), this documentation
update should save users some pain when trying to override or
disable the default entrypoint for a container.


jeremybumsted (approved):
LGTM, thank you for adding this @bunsenmcdubbs !

PR #194, opened 2022-02-11
fix YAML so renovate doesn't complain about invalid configs

jeremybumsted:
This PR is to address #110 

This should _hopefully_ bring the config in line with the [Renovate docs](https://docs.renovatebot.com/configuration-options/#labels)


pzeballos (approved):
LGTM!

PR #202, opened 2022-06-06
Allow switching '--interactive' off

laurynaslubys:
Many build jobs are not interactive and "hang" if stdin is connected. Allow disabling the `--interactive` flag that is passed in to docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ steps:
     artifact_paths: "dist/**"
 ```
 
-You can pass in additional environment variables and customize what is mounted into the container. 
+You can pass in additional environment variables and customize what is mounted into the container.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -287,6 +287,10 @@ If set to false, doesn't allocate a TTY. This is useful in some situations where
 
 Default: `true` for Linux and macOS, and `false` for Windows.
 
+### `interactive` (optional, boolean)
+
+If set to false, doesn't connect `stdin` to the process. Some scripts fall back to asking for user input in case of errors, if the process has `stdin` connected and this results in the process waiting for input indefinitely.
+
 ### `user` (optional, string)
```
motiejus, line 294: Not clear which one is the default. Looking at the code it seems like it's false.

In that case, a clearer wording would be "If set to true, connect to `stdin` of the process" or so.

laurynaslubys: The default is true, so the wording is probably correct. The wording is based on a sibling parameter `tty`

toote:
I have added the tests missing on this PR and corrected all the ones it broke (basically all of them :P). While I was at it, I found that the option to remove the `-t` from the command was also not being tested so I added that one as well

pzeballos (approved):
🙌🏻

PR #206, opened 2022-09-01
Add skip-checkout option

tetienne:
I often meet the case where all the files I need I already contained within the Docker image. In such case, there is no reason to checkout the repository.
The docker-compose plugin already implements this [option](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin#skip-checkout-optional-run-only).

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,9 +275,11 @@ Set the size of the `/dev/shm` shared memory filesystem mount inside the docker 
 
 Example: `2gb`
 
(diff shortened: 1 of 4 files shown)


toote:
Added tests to the PR and corrected a bug with the implementation :)

tetienne:
Thx for the update

pzeballos (approved):
🎉

PR #171, opened 2020-11-13
Add options for --cap-add, --cap-drop, and --security-opt

christopher-dG:
Adding some missing options for `docker run` 🙂

PR #207, opened 2022-09-14
Add options for --cap-add, --cap-drop, and --security-opt

toote:
This is the exact same thing as #171 but added missing tests (which could not be done as the original fork no longer exists). I was able, though, to keep the original commit intact to respect to the original author.


pzeballos (approved):
This looks good! 🎉

PR #208, opened 2022-09-15
Fix plugins (and `--entrypoint` behaviour)

toote:
This fixes the 2 tests that started failing since the release of the latest plugin tester.

That uncovered a bug in the tests themselves of the `--entrypoint` option which is what actually caused #138. The easiest solution to which was to just remove the alternative of the plugin option to be a boolean. It needs to be a string that will be used as-is and updated the documentation accordingly.

Due to the complex interaction between entrypoint, shell and command as defined in this plugin, the docker container and the pipeline that actually used the plugin I left the reference to the issue in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ steps:
   - command: "go build -o dist/my-app ."
     artifact_paths: "./dist/my-app"
     plugins:
-      - docker#v3.13.0:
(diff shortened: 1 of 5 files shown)


pzeballos (approved):
🎉🎉

PR #211, opened 2022-09-20
Change `mount-buildkite-agent` default

### LINKED ISSUE #149: mount-buildkite-agent being `true` by default is a footgun
We're using the docker-buildkite-plugin to achieve a basic level of build isolation to make it safer to run pull requests from third-party forked repositories.

However, for this to be effective, we need to make sure we set `mount-buildkite-agent: false` for each invocation of the docker-buildkite-plugin, as otherwise the agent token is leaked into the docker container, allowing a malicious PR author to run arbitrary pipelines on our build agents.

I'd like to suggest that `mount-buildkite-agent` should be false by default.

### LINKED ISSUE #154: mount-buildkite-agent doesn't work across platforms
As mentioned in https://github.com/buildkite/agent/issues/1181, mount-buildkite-agent doesn't work when running a buildkite agent on a macos host and using something like docker for mac because it mounts a mac binary into a linux container.

There are a couple of options, we could at least raise a clear error message, or we could try to provision and make available a target platform-appropriate binary.

### LINKED ISSUE #157: mount-buildkite-agent on Windows fails with invalid volume specification
```txt
C:\Program Files\docker\docker.exe: Error response from daemon: invalid volume specification: '/c/program files/buildkite-agent/buildkite-agent:/usr/bin/buildkite-agent'.
```

Plugin version v3.5.0.

Windows 2019 Server.

toote:
This changes the default for the option and adds a warning in the readme associated to the option.

**Important**: this means that this should be a major release as it changes default behaviour

Closes #149
Closes #157 
Closes #154

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ steps:
   - command: "go build -o dist/my-app ."
     artifact_paths: "./dist/my-app"
     plugins:
-      - docker#v4.0.0:
+      - docker#v4.1.0:
           image: "golang:1.11"
 ```
 
@@ -23,7 +23,7 @@ Windows images are also supported:
 steps:
   - command: "dotnet publish -c Release -o published"
     plugins:
-      - docker#v4.0.0:
+      - docker#v4.1.0:
           image: "microsoft/dotnet:latest"
           always-pull: true
 ```
@@ -33,7 +33,7 @@ If you want to control how your command is passed to the docker container, you c
 ```yml
 steps:
   - plugins:
-      - docker#v4.0.0:
+      - docker#v4.1.0:
           image: "mesosphere/aws-cli"
           always-pull: true
           command: ["s3", "sync", "s3://my-bucket/dist/", "/app/dist"]
@@ -50,7 +50,7 @@ steps:
       - "yarn install"
       - "yarn run test"
     plugins:
-      - docker#v4.0.0:
+      - docker#v4.1.0:
           image: "node:7"
           always-pull: true
           environment:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -219,9 +219,13 @@ Default: `true`
 
 ### `mount-buildkite-agent` (optional, boolean)
 
-Whether to automatically mount the `buildkite-agent` binary from the host agent machine into the container. Set to `false` if you want to disable, or if you already have your own binary in the image.
+Whether to automatically mount the `buildkite-agent` binary from the host agent machine into the container.
 
-Default: `true` for Linux, and `false` for macOS and Windows.
+Set to `true` if you want to enable and are sure that the binary running in the agent is compatible with the container's architecture and environment (for example, don't try to mount the OS X or Widnows agent binary in a container running linux). If enabled in Windows agents your pipeline, step or agent **must have the `BUILDKITE_AGENT_BINARY_PATH` environment variable defined** with the executable to mount in the (Windows) agent.
```
pzeballos, line 224:
```suggestion
Set to `true` if you want to enable and are sure that the binary running in the agent is compatible with the container's architecture and environment (for example, don't try to mount the OS X or Windows agent binary in a container running linux). If enabled in Windows agents your pipeline, step or agent **must have the `BUILDKITE_AGENT_BINARY_PATH` environment variable defined** with the executable to mount in the (Windows) agent.
```

pzeballos (changes requested):
It will be necessary to update the docker version in the README for 4.2.1 (current version)

pzeballos (approved):

PR #217, opened 2022-09-24
Volume variable interpolation

### LINKED ISSUE #189: Can't interpolate ENV vars in mount path
_Note: This seems related to #185, but the workaround suggested does not apply for my use case._

I just submitted buildkite-plugins/ecr-buildkite-plugin#72, but I'm not sure where the issue lies. 

TL;DR: I'm trying to mount my `${DOCKER_CONFIG}` folder into a docker container, but it appears the `ecr` plugin isn't setting it in such a way that I can't simply put `${DOCKER_CONFIG}:/home/user/.docker` in the `volumes` array.

Per the [docs](https://buildkite.com/docs/pipelines/secrets#anti-pattern-referencing-secrets-in-your-pipeline-yaml), however, it seems I _should_ be able to add an extra dollar sign (so, `$${DOCKER_CONFIG}:/home/user/.docker`) and make it work. However, for whatever reason, although `$${DOCKER_CONFIG}` is converted to `${DOCKER_CONFIG}`, it isn't expanded beyond that, and the literal string `${DOCKER_CONFIG}` is sent to `docker` as an argument.

Would it be possible to turn on interpolation for the volume paths so that this works?

toote:
Adds a new option to allow for variable interpolation in the volumes option.

It is behind a gate because it uses `eval` so it may be a security risk.

Closes #189

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,6 +329,12 @@ Extra volume mounts to pass to the docker container, in an array. Items are spec
 
 Example: `[ "/var/run/docker.sock:/var/run/docker.sock" ]`
 
+### `expand-volume-vars` (optional, boolean, run only, unsafe)
+
+When set to true, it will activate interpolation of variables in the elements of the `volumes` configuration array. When turned off (the default), attempting to use variables will fail as the literal `$VARIABLE_NAME` string will be passed to the `-v` option.
+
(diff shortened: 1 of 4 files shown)


pzeballos (approved):
💚

PR #220, opened 2022-09-28
Pass ECS env vars when using propagate-aws-auth-tokens

francoiscampbell:
When the agent is running in ECS, credentials are obtained by the various AWS SDKs by contacting `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`. This PR passes this env var from the agent to the docker container if it's defined.

Update: Also passes `AWS_CONTAINER_CREDENTIALS_FULL_URI` and `AWS_CONTAINER_AUTHORIZATION_TOKEN`

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ steps:
           propagate-environment: true
 ```
 
-AWS authentication tokens can be automatically propagated to the container, for example from an assume role plugin:
+AWS authentication tokens can be automatically propagated to the container, for example from an assume role plugin or ECS IAM role:
 
 ```yml
 steps:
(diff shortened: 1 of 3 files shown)


toote:
That is actually a great change, but based on [the corresponding AWS documentation](https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html) I would suggest you add all 3 variables necessary for working in ECS. What do you think?

francoiscampbell:
I've never had to pass `AWS_CONTAINER_AUTHORIZATION_TOKEN` when working with agents in ECS before, and `AWS_CONTAINER_CREDENTIALS_FULL_URI` isn't necessary if `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is set:

> Note: This setting is an alternative to AWS_CONTAINER_CREDENTIALS_RELATIVE_URI and will only be used if AWS_CONTAINER_CREDENTIALS_RELATIVE_URI is not set in the same profile.

but since this only takes effect when they're set in the external env, I don't mind passing all 3.

toote (approved):

Issue #184, opened 2021-05-10
Why doesn't `propagate-environment` include env vars set in hooks?


evandam:
Hi folks,

Just wondering what the reason/limitation is behind why setting `propagate-environment: true` does not include environment variables that are set by hooks. The current behavior is clearly documented, but I'm unsure of why since it seems like it would be a very common and useful feature to have.

For example, I would expect the following to set `GIT_SHA` correctly:

`/etc/buildkite-agent/pre-command`:

```bash
export GIT_SHA="$(git rev-parse --short HEAD)"
```

`pipeline.yml`:

```yaml
steps:
  - command: "echo $$GIT_SHA"
    env:
      FOO: bar
    plugins:
      - docker#v3.8.0:
          image: busybox
          propagate-environment: true
```

I need to pass `environment: ["GIT_SHA"]` explicitly in the plugin config for it to work. It would be helpful to include these environment variables, since a very common use of hooks is to expose environment variables to a step. It would also be very helpful in case a lot of env vars are set in hooks, and the steps don't need to keep the list in sync.

pzeballos:
Hi @evandam,


> For example, I would expect the following to set GIT_SHA correctly:
> /etc/buildkite-agent/pre-command:

I think the agent hook directory should be: `/etc/buildkite-agent/hooks/pre-command`

Can you confirm that the `pre-command` hook is definitely running? and if that is the case, could you send us some log output to confirm that everything is ok?

pzeballos:
Hi @evandam,

Sorry for the delay; we were discussing your issue with the rest of the team.
What you point out makes sense, and it would be logical to include environment variables exported from preceding hooks.
It only propagates the standard list of job environment variables, and ones defined in the pipeline.yml, because it uses [$BUILDKITE_ENV_FILE](https://buildkite.com/docs/pipelines/environment-variables#bk-env-vars-buildkite-env-file):

https://github.com/buildkite-plugins/docker-buildkite-plugin/blob/master/hooks/command#L262-L264

But this sounds like an excellent suggestion, and it would be a great feature enhancement if you like to submit a PR, we would love that!

Cheers!

evandam:
Hi @pzeballos no worries! I don't think I'm familiar enough with the Buildkite internals to be the one to open a PR 😅 

[...]

PR #238, opened 2023-02-26
`env-propagation-list`: programmatically propagate envvars to the container

edmund-huber:
Hi! Like many others, https://github.com/buildkite-plugins/docker-buildkite-plugin/issues/184 , I tried setting env vars in a hook, then was surprised when `propagate-environment` didn't propagate them to the container.

Here is my proposal for how to fix this. Instead of changing the behavior of `propagate-environment`, add a new option `env-propagation-list`, which:
> If you set this to `VALUE`, and `VALUE` is an environment variable containing a space-separated list of environment variables such as `A B C D`, then A, B, C, and D will all be propagated to the container. This is helpful when you've set up an `environment` hook to export secrets as environment variables, and you'd also like to programmatically ensure that secrets get propagated to containers, instead of listing them all out.

Here's how to use it: in your `environment` hook, where you would normally be generating bash statements like:
```
export SECRET_1="one"
export SECRET_2="two"
```
You can instead do:
```
export SECRET_1="one"
export PROPAGATE_LIST="SECRET_1 ${PROPAGATE_LIST:-}"
export SECRET_2="two"
export PROPAGATE_LIST="SECRET_2 ${PROPAGATE_LIST:-}"
```
Then, when you invoke this plugin, give it `propagate-list: 'PROPAGATE_LIST'`, and those vars you set will end up in the container, too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,11 +177,17 @@ Example: `[ "BUILDKITE_MESSAGE", "MY_SECRET_KEY", "MY_SPECIAL_BUT_PUBLIC_VALUE=k
 
 An array of additional files to pass into to the docker container as environment variables. Each entry corresponds to a Docker CLI `--env-file` parameter.
 
+### `env-propagation-list` (optional, string)
+
+If you set this to `VALUE`, and `VALUE` is an environment variable containing a space-separated list of environment variables such as `A B C D`, then A, B, C, and D will all be propagated to the container. This is helpful when you've set up an `environment` hook to export secrets as environment variables, and you'd also like to programmatically ensure that secrets get propagated to containers, instead of listing them all out.
+
 ### `propagate-environment` (optional, boolean)
 
-Whether or not to automatically propagate all pipeline environment variables into the docker container. Avoiding the need to be specified with `environment`.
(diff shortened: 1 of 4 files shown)


edmund-huber:
I forgot to mention: I'm opening this PR because listing envvars you want to propagate manually in every step that uses docker-plugin and needs envvars is pretty annoying and error-prone.

toote (changes requested):
This is a very interesting feature!

The one thing I would change is to define, implement, document and test the behaviour of when the configured variable does not exist. Right now, the execution explodes with:

```
/plugin/hooks/command: line 316: !BUILDKITE_PLUGIN_DOCKER_ENV_PROPAGATION_LIST: unbound variable
```

That I think is less than ideal

edmund-huber:
How's this?

toote (approved):

amartani:
I opened https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/388 to implement this feature for the docker-compose plugin, as it would be useful there as well. Thanks!

PR #239, opened 2023-03-04
add `arbitrary-flags`, which runs a given arbitrary script in order to generate more flags to `docker run`

edmund-huber:
For my workload, I happen to need to run docker like `docker run --group-add=$(getent group docker | cut -d: -f3)`. I cannot do exactly that with this plugin currently, and have to resort to `docker run --group-add=997` which feels bad because the number is pretty inscrutible, also it's fragile.

So I'm opening this PR to start a discussion of how we can allow arbitrary flags generated at runtime. What do you think about this or something like this, allowing the user to define at runtime what args `docker run` is going to get?

### DIFF
--- a/hooks/command
+++ b/hooks/command
@@ -529,6 +529,14 @@ elif [[ ${#command[@]} -gt 0 ]] ; then
   done
 fi
 
+# Support running an arbitrary shell script to generate a list of additional
+# options to pass to docker.
+arbitrary_script="${BUILDKITE_PLUGIN_DOCKER_ARBITRARY_FLAGS:-}"
+arbitrary_args=""
+if [[ -n "$arbitrary_script" ]] ; then
+  arbitrary_args=$(bash -c "$arbitrary_script")
+fi
+
 echo "--- :docker: Running command in ${BUILDKITE_PLUGIN_DOCKER_IMAGE}"
 echo -ne '\033[90m$\033[0m docker run ' >&2
 
@@ -542,7 +550,7 @@ set +e
 
 # Don't convert paths on gitbash on windows, as that can mangle user paths and cmd options.
 # See https://github.com/buildkite-plugins/docker-buildkite-plugin/issues/81 for more information.
-( if is_windows ; then export MSYS_NO_PATHCONV=1; fi && docker run "${args[@]}" )
+( if is_windows ; then export MSYS_NO_PATHCONV=1; fi && docker run $arbitrary_args "${args[@]}" )
 
 exit_code=$?
 
--- a/plugin.yml
+++ b/plugin.yml
@@ -97,6 +97,8 @@ configuration:
       type: array
     ulimits:
       type: array
+    arbitrary_flags:
+      type: string
   required:
     - image
   additionalProperties: false
--- a/tests/command.bats
+++ b/tests/command.bats
@@ -1227,3 +1227,18 @@ EOF
 
   unstub docker
 }
+
+@test "Runs BUILDKITE_COMMAND with a set of arbitrary flags" {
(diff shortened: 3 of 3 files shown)


toote:
There are several issues with doing something as what you are suggesting. The main one is that [`shellcheck` is complaining about the way you are using the variable](https://buildkite.com/buildkite/plugins-docker/builds/198#0186a9f1-79f0-43e3-a1ac-ff72adba8a82/93-96), and correcting that will break your use-case unless you are adding a single option that either takes no parameter or is used in the `--option=value` format... but that would kind of dewfeat the whole purpose of the option. Another one I can think of is the lack of control that would provide on conflicting options. Finally, I'm not sure I'm comfortable with the security implications of allowing random scripts determine what this plugin do.

In your particular case, why can't you just use the option `additional-groups: docker` instead? There is no need to run additional commands to get the id of the group as the [`docker run` command can assign them off names themselves](https://docs.docker.com/engine/reference/run/#additional-groups). If that is not something you believe is good enough, another possible solution would be to have an environment hook (either on the agent or the repository) that creates a variable with the corresponding value that you can either use when uploading the pipeline. Or at least that is what I could come up with with the limited information on your scenario :)

edmund-huber:
Yeah, in this case, shellcheck is wrong.

> Finally, I'm not sure I'm comfortable with the security implications of allowing random scripts determine what this plugin do.
I guess this is true in the sense that anyone could add that env var? Like, the fact that `BUILDKITE_PLUGIN_DOCKER_ARBITRARY_FLAGS` exists doesn't mean that it was a plugin option in the buildkite yaml, it could also have been any random program assigning it.

`additional-groups: docker` doesn't work in my case because the uid/gid mapping in the host is different than that of the container.

> another possible solution would be to have an environment hook (either on the agent or the repository) that creates a variable with the corresponding value that you can either use when uploading the pipeline.

As in, pass `additional-groups: $ENV_VAR` in the plugin configuration? I'm not sure I understand.

I'll probably just keep my fork that allows arbitrary flags, because it's just too convenient. Thanks!

toote:
> Yeah, in this case, shellcheck is wrong.

Actually, it is a valid error because such constructs would allow for command injection attacks.

> > Finally, I'm not sure I'm comfortable with the security implications of allowing random scripts determine what this plugin do.
> > I guess this is true in the sense that anyone could add that env var? Like, the fact that `BUILDKITE_PLUGIN_DOCKER_ARBITRARY_FLAGS` exists doesn't mean that it was a plugin option in the buildkite yaml, it could also have been any random program assigning it.
> 

That, what I previously mentioned, and the fact that even if you trust the plugin itself, a value here could execute arbitrary code... for example, code that downloads a script off the internet and runs more stuff, or shares some secret file or environment variable if anyone can set the correct variable in that particular step. Of course, that may be too paranoid of me; but I'd rather not provide any means for people to hammer their own fingers.

> `additional-groups: docker` doesn't work in my case because the uid/gid mapping in the host is different than that of the container.

It is possible you can fix those with [docker's user namespace remapping](https://docs.docker.com/engine/security/userns-remap/), but it may be more headaches than those you already have :(

> > another possible solution would be to have an environment hook (either on the agent or the repository) that creates a variable with the corresponding value that you can either use when uploading the pipeline.
> 
> As in, pass `additional-groups: $ENV_VAR` in the plugin configuration? I'm not sure I understand.

Yup... you have a a hook that exports the docker group id as an environment variable and then use that id when you do the pipeline upload. So you still get the benefits of using a standard option of the plugin and defining the ID programatically.

It is important to note that the variable will be interpolated during pipeline upload so it has to be available at that point and its value be the same as the agent that will end up running the corresponding step.

PR #242, opened 2023-03-14
Support for --platform arg

jpfielding:
use case: i want to run a cluster of mostly linux/amd64, but i need to test on arm too.  rolling up an agent just for that is more work than just running on an arm capable container.

### DIFF
--- a/tests/command.bats
+++ b/tests/command.bats
@@ -713,9 +713,6 @@ EOF
   export BUILDKITE_PLUGIN_DOCKER_PLATFORM=linux/${DOCKER_ARCH}
   export BUILDKITE_COMMAND="echo hello world"
 
-  # peaking in on docker to see
-  echo $(docker --version)
-
   stub docker \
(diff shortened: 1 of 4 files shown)


tests/command.bats
```diff
@@ -708,6 +708,22 @@ EOF
   unstub docker
 }
 
+@test "Runs BUILDKITE_COMMAND with platform" {
+  export DOCKER_ARCH=$(arch | sed 's/aarch64/arm64/' | sed 's/x86_64/amd64/')
```
toote, line 712: there is no need to inspect the actual architecture as the command will never be actually executed, you can fix this to whatever value you want as long as it is the same in the stub

toote (changes requested):
great addition! Just a minor comment to simplify the test :)

tests/command.bats
jpfielding: ah ok, i wasnt sure if it was going to run, and wanted to be sure it was safe wherever it ran, but i can hard code that. 

thanks!

toote (approved):

PR #249, opened 2023-05-23
feat: Add AWS Web Idenity compability to PROPAGATE_AWS_AUTH_TOKENS

zopanix:
This adds compatibilty with AWS authentication using delegated authentication using web identity token files (useful with AWS EKS for example). It adds an environment variable and a volume if the AWS_WEB_IDENTITY_TOKEN_FILE is present to be able to use them from the specified container.

In addition, we have added the AWS_STS_REGION_ENDPOINTS variable compatibility which allows you AWS IAM calls to be directed towards the region you are in instead of redirecting automatically to us-east-1 by default.

If you need more explanation or have concerns with the code, I'll be happy to address and/or correct them if needed :)


toote (approved):

toote:
Thanks a lot for the contribution!

Issue #253, opened 2023-06-19
Support for other docker commands than docker run


linus345:
Hey, are there any plans to support other docker commands then running a command in a container? For example introducing support for `docker load`. This would be useful in cases where you want to load a docker image before running a command in a container.

If there's no plans to add this, would you accept a PR introducing it if I opened one up?

toote:
That is quite an interesting idea and we would be more than happy to review a PR if you create one.

Note, though that running a `docker load` on its own without an actual command for the plugin to take care might not make sense. It is possible that the clarification is not necessary, but I believe it is necessary because `command` is a term that has many meanings in this context (`docker` cli subcommand, command hook and actual step's command). I think that if you create a PR explaining the use case and proper tests we can continue the discussion there as well :)

linus345:
Created a PR for this #254, would appreciate any feedback you have

PR #254, opened 2023-06-20
Add support for `docker load`

linus345:
This PR is an attempt to implement support for `docker load`.

Related issue #253.

It first makes a small refactor by moving out the `run` command code into its own file and making the `hooks/command` file only the entry point. This is the same structure that [buildkite-plugins/docker-compose-buildkite-plugin](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin) uses. It then introduces a new file for the `docker load` command.


linus345:
The example added in the README file is the use case that I was thinking of and that we need. I'm not sure if it makes sense to use `load` without running a container afterwards though.

hooks/command
```diff
@@ -1,575 +1,23 @@
...
-fi
-
-# Parse publish args and add them to docker args
-if plugin_read_list_into_result BUILDKITE_PLUGIN_DOCKER_PUBLISH ; then
-  for arg in "${result[@]}" ; do
-    args+=( "--publish" "$arg" )
-  done
+if in_array "LOAD" "${commands[@]}" ; then
+  # shellcheck source=commands/load.sh
+  . "$DIR/../commands/load.sh"
```
toote, line 16: why source the script when you could just run it? (note that you will have to load the `shared.bash` library in the script as well if making the change)

hooks/command
```diff
@@ -1,575 +1,23 @@
...
 fi
 
-if [[ -n "${BUILDKITE_PLUGIN_DOCKER_PROPAGATE_UID_GID:-}" ]] ; then
-  args+=("-u" "$(id -u):$(id -g)")
+# if the `image` parameter is supplied we imply `docker run` to not break backwards compatibility by introducing a new `run` parameter
+if in_array "IMAGE" "${commands[@]}" ; then
+  # shellcheck source=commands/run.sh
+  . "$DIR/../commands/run.sh"
```
toote, line 22: same here, not sure why sourcing the command when you could just run it (and, once more, changing this would mean loading the `shared.bash` in that script as well

hooks/command
```diff
@@ -1,575 +1,23 @@
...
-# Returns success if values were read
-plugin_read_list_into_result() {
-  result=()
+commands=()
 
-  for prefix in "$@" ; do
-    local i=0
-    local parameter="${prefix}_${i}"
+[[ -n "$(plugin_read_list LOAD)" ]] && commands+=("LOAD")
+[[ -n "$(plugin_read_list IMAGE)" ]] && commands+=("IMAGE")
```
toote, line 12: if you follow my suggestion (about the loading being conditional but not the run portion) this array and subsequent logic would not be necessary

lib/shared.bash
```diff
@@ -0,0 +1,121 @@
...
+  else
+    echo "$path"
+  fi
+}
+
+function in_array() {
+  local e
+  for e in "${@:2}"; do [[ "$e" == "$1" ]] && return 0; done
+  return 1
+}
```
toote, line 110: if you follow my suggestion removing the `command` array from the main hook , this function would not be necessary

tests/command.bats
```diff
@@ -1306,3 +1306,53 @@ EOF
 
   unstub docker
 }
+
+export BUILDKITE_AGENT_STUB_DEBUG=/dev/tty
```
toote, line 1310: this should be part of the comment at the top of the file (and commented out :P)

toote (commented):
This is a great PR! I really appreciate your splitting the huge command hook into separate files.

That said, I don't see the use-case for loading an image without actually running anything on it. The very example you show of using the `load` configuration also contains an `image` tag. And if it didn't, I'm not sure using the plugin would make sense in the first place as the docker daemon running that load step may very well not be used for anything else so making the `run` portion of the plugin inconditional anyways.

I would suggest that the `load.sh` is executed conditionally as it is now, but the run should always run no matter what.

What do you think?

hooks/command
linus345: The reason for sourcing it instead of directly executing the script is just because I was keeping it similar to the structure of the docker-compose plugin where it's done the same way [here](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/master/hooks/command#L31).

linus345:
Yes I agree with that in the current state, I'll make the run required. Although I do think it can make sense in the future if other commands like `push` are implemented.

hooks/command
linus345: Updated according to your comments, except this about sourcing because of the reason above. Let me know if you would rather have the script being ran instead of sourced and I'll change it.

commands/load.sh
```diff
@@ -0,0 +1,21 @@
...
+arg=""
+
+if plugin_read_value_into_result BUILDKITE_PLUGIN_DOCKER_LOAD ; then
+  arg="${result[0]}"
+fi
+
+if [[ -z "$arg" ]]; then
+  echo "Error: load has to be given a path to the file that should be loaded"
+  exit 1
+fi
```
toote, line 14: According to the plugin spec, the `load` option is just a string so it makes a lot more sense to just do:

```suggestion
arg="${BUILDKITE_PLUGIN_DOCKER_LOAD:-}"

if [ -z "$arg" ] ; then
  echo "Error: load has to be given a path to the file that should be loaded"
  exit 1
fi
```

commands/load.sh
```diff
@@ -0,0 +1,21 @@
...
+
+echo "load arg: $arg"
```
toote, line 16: I'm not sure this is necessary... but if you want to print out some message, I would make it more user friendly:

```suggestion
echo "Loading image from $arg"
```

hooks/command
```diff
@@ -1,589 +1,16 @@
...
-
-# Handle setting of shm size if provided
-if [[ -n "${BUILDKITE_PLUGIN_DOCKER_SHM_SIZE:-}" ]]; then
-  args+=("--shm-size" "${BUILDKITE_PLUGIN_DOCKER_SHM_SIZE}")
-fi
-
-# Handle setting of cpus if provided
-if [[ -n "${BUILDKITE_PLUGIN_DOCKER_CPUS:-}" ]]; then
-  args+=("--cpus=${BUILDKITE_PLUGIN_DOCKER_CPUS}")
+if [[ -n "$(plugin_read_list LOAD)" ]]; then
```
toote, line 9: Here the same, as the `load` option is just a string, you should just instead use:

```suggestion
if [[ -n "${BUILDKITE_PLUGIN_DOCKER_LOAD:-}" ]]; then
```

tests/command.bats
```diff
@@ -1342,3 +1342,51 @@ EOF
...
+  stub docker \
+    "load -i test-image.tar.gz : echo loaded docker image"
+
+  run "$PWD"/hooks/command
+
+  assert_success
+  assert_output --partial "loaded docker image"
+
+  unstub docker
+}
```
toote, line 1374: These two tests are failing because now the `run` part of the plugin is unconditional. I believe they should be removed anyways.

toote (changes requested):
Just a few minor corrections. While I was at it, I also noticed that you added a `load` example to the plugin but did not document the configuration itself in the `Optional` section of README.md (make sure to keep the alphabetical order of them when you add it)

linus345:
Thanks for the feedback. Fixed the comments now.

commands/run.sh
```diff
@@ -0,0 +1,542 @@
...
+
+if [[ "${BUILDKITE_PLUGIN_DOCKER_ALWAYS_PULL:-false}" =~ ^(true|on|1)$ ]] ; then
+  echo "--- :docker: Pulling ${image}"
+  if ! retry "${BUILDKITE_PLUGIN_DOCKER_PULL_RETRIES:-3}" \
+       docker pull "${image}" ; then
+    echo "!!! :docker: Pull failed."
+    # retry_exit_status is set in the retry function, don't warn about it being unassigned
+    # shellcheck disable=SC2154
+    exit "$retry_exit_status"
+  fi
```
toote, line 342: A better way to do this without disabling shellcheck nor getting a warning/error is using the fact that the `retry` fuction actually `returns` the value to use so it is assigned to its exit code:

```suggestion
  if ! retry "${BUILDKITE_PLUGIN_DOCKER_PULL_RETRIES:-3}" \
       docker pull "${image}" ; then
    retry_exit_status="$?"
    echo "!!! :docker: Pull failed."
    exit "$retry_exit_status"
  fi
```

I tested this myself:

```
$ function test_return { local exit_code=$1; return $exit_code; }
$ test_return 1
$ echo $?
1
$ test_return 0
$ echo $?
0
$ if test_return 0; then echo "OK: $?"; else echo "FAIL: $?"; fi
OK: 0
$ if test_return 234; then echo "OK: $?"; else echo "FAIL: $?"; fi
FAIL: 234
```

.buildkite/pipeline.yml
```diff
@@ -2,7 +2,13 @@ steps:
   - label: ":shell: Shellcheck"
     plugins:
       - shellcheck#v1.3.0:
-          files: hooks/**
+          files:
+            - hooks/**
+            - commands/**
+            - libs/**
+          options:
+            - "--color=always"
```
toote, line 10: this is not necessary, it is [always turned on in the plugin](https://github.com/buildkite-plugins/shellcheck-buildkite-plugin/blob/50eda5bb57896d234408d7dcaa9fccc15ff6e953/hooks/command#L99)

.buildkite/pipeline.yml
```diff
@@ -2,7 +2,13 @@ steps:
   - label: ":shell: Shellcheck"
     plugins:
       - shellcheck#v1.3.0:
-          files: hooks/**
+          files:
+            - hooks/**
+            - commands/**
+            - libs/**
```
toote, line 8: There is a typo here, it should be:
```suggestion
            - lib/**
```

.buildkite/pipeline.yml
```diff
@@ -2,7 +2,13 @@ steps:
...
+          options:
+            - "--color=always"
+            - "--external-sources"
```
toote, line 11: After correcting the typo above, this is no longer necessary

commands/run.sh
```diff
@@ -0,0 +1,542 @@
...
+done < <(env | sort)
+
+# Privileged container
+if [[ "${BUILDKITE_PLUGIN_DOCKER_PRIVILEGED:-false}" =~ ^(true|on|1)$ ]] ; then
+    args+=( "--privileged" )
+fi
+
+if plugin_read_list_into_result BUILDKITE_PLUGIN_DOCKER_ENV_FILE; then
+  # don't warn about result being unassigned (plugin_read_list_into_result assigns to global result variable)
+  # shellcheck disable=SC2154
```
toote, line 254: All these disabling of checks can be easily avoided if you just source the shared library that defines `plugin_read_list_into_result` at the top:

```
DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"

# shellcheck source=lib/shared.bash
. "$DIR/../lib/shared.bash"
```

toote (changes requested):
Good catch with the shellcheck plugin not including all the new files! :clap: 

I tend to avoid disabling checks so I provided alternatives for all such instances (plus found that the additional options are not necessary due to a typo)

.buildkite/pipeline.yml
linus345: From the testing I did it seems that this is still needed. Shellcheck won't follow the source directives if `--external-sources` is not used.

linus345:
Finally got around to fixing the comments.

.buildkite/pipeline.yml
toote: That is really weird, I ran the pipeline without the option and shellcheck did not fail. Can you try removing it so that we can further debug the issue seeing an actual run?

pzeballos: Hey @linus345! thanks again for this! are you able to try toote's suggestions so we can see the issue? thanks!

linus345: Completely missed that this comment was made.

Sure I'll try it out tomorrow.

linus345: Sorry for taking such a long time, pushed it without this commit now.

toote (approved):
:clap: sorry for the delay in getting back around to this

PR #255, opened 2023-07-05
Add Docker labels to container

francoiscampbell:
Same changes as https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/376. This is useful to us since we can map labels to Datadog tags and track metrics and usage from the containers spawned by the agent using this plugin.

Based on the discussion in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/376#issuecomment-1449177002, I've made the option default to `true`, since it's unlikely to break anything, but we still want to give the option to turn it off.

Tested on our infra:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,6 +293,12 @@ Specify the IPC mode to use. See the [docker run options documentation](https://
 
 Example: `host`
 
+### `run-labels` (optional, boolean)
+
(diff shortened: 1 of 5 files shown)


pzeballos (approved):
Thanks @francoiscampbell! This looks great! 🎉

PR #269, opened 2024-04-09
feat: add GCP workload identity federation compatibility

valkum:
We run our workloads in Google Cloud and thus need to identify with a workload identity provider. I noticed for the AWS counterpart there is special support with `propagate-aws-auth-tokens`.

This adds support for the `gcp-workload-identity-federation` plugin and works similar to the AWS token support.
It will expose the env vars set by `gcp-workload-identity-federation` and mount the OIDC temp dir created by the plugin.

I wonder if in the future these can be more abstracted to avoid adding a setting for each cloud provider.


pzeballos (approved):
Thanks @valkum! This looks good! We'll do a release next week.

>I wonder if in the future these can be more abstracted to avoid adding a setting for each cloud provider.

Yeah, we would like that and it's in the backlog but we don't have a timeframe for this

PR #278, opened 2024-10-07
Add `chown` setting to fix root permissions on exit

HackAttack:
Implementation borrowed from [TeamCity](https://www.jetbrains.com/help/teamcity/cloud/container-wrapper.html#Restoring+File+Ownership+on+Linux)

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ Will propagate `GOOGLE_APPLICATION_CREDENTIALS`, `CLOUDSDK_AUTH_CREDENTIAL_FILE_
 
 Whether to match the user ID and group ID for the container user to the user ID and group ID for the host user. It is similar to specifying `user: 1000:1000`, except it avoids hardcoding a particular user/group ID.
 
-Using this option ensures that any files created on shared mounts from within the container will be accessible to the host user. It is otherwise common to accidentally create root-owned files that Buildkite will be unable to remove, since containers by default run as the root user.
+Using this option ensures that any files created on shared mounts from within the container will be accessible to the host user. It is otherwise common to accidentally create root-owned files that the agent will be unable to remove, since a lot of images by default run as the root user. This can also be mitigated for the checkout directory by using the `chown` option of this plugin.
 
 ### `privileged` (optional, boolean)
 
@@ -318,6 +318,20 @@ Whether to automatically mount the current working directory which contains your
 If there's a git mirror path and `mount-checkout` is enabled, the (mirror path)[https://buildkite.com/docs/pipelines/environment-variables#BUILDKITE_REPO_MIRROR] is mounted into the docker container as an added volume. Otherwise, the git mirror path will have to be explicitly added as an extra volume to mount into the container. 
 
 Default: `true`
+
+### `chown` (optional, boolean)
+
+Whether to `chown` the directory mounted with `mount-checkout` to the Buildkite agent user before exiting, to ensure that the agent can clean up any additional files created there. This will happen even if the command fails or is cancelled. This option has no effect on Windows or if `mount-checkout` is false.
+
+Prefer using `propagate-uid-gid` over this option, as the `chown`–which can take some time if your checkout is of considerable size—is likely not needed at all in that case.
+
+Default: `false`
+
+### `chown-image` (optional, string)
+
+The Docker image in which to run the `chown` command.
+
+Default: `busybox`
 
 ### `mount-buildkite-agent` (optional, boolean)
 
--- a/hooks/pre-exit
+++ b/hooks/pre-exit
@@ -1,10 +1,19 @@
 #!/bin/bash
 set -euo pipefail
 
+DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"
+
+# shellcheck source=lib/shared.bash
+. "$DIR/../lib/shared.bash"
+
 if [[ "${BUILDKITE_PLUGIN_DOCKER_CLEANUP:-true}" =~ ^(true|on|1)$ ]] ; then
   for container in $(docker ps -a -q --filter "label=com.buildkite.job-id=${BUILDKITE_JOB_ID}") ; do
     echo "~~~ Cleaning up left-over container ${container}"
     docker stop "$container"
   done
+fi
+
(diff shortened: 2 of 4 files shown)


hooks/pre-exit
```diff
@@ -1,10 +1,19 @@
...
+. "$DIR/../lib/shared.bash"
+
 if [[ "${BUILDKITE_PLUGIN_DOCKER_CLEANUP:-true}" =~ ^(true|on|1)$ ]] ; then
   for container in $(docker ps -a -q --filter "label=com.buildkite.job-id=${BUILDKITE_JOB_ID}") ; do
     echo "~~~ Cleaning up left-over container ${container}"
     docker stop "$container"
   done
 fi
 
+if ! is_windows && [[ "${BUILDKITE_PLUGIN_DOCKER_MOUNT_CHECKOUT:-on}" =~ ^(true|on|1)$ ]] && [[ "${BUILDKITE_PLUGIN_DOCKER_CHOWN:-false}" =~ ^(true|on|1)$ ]] ; then
```
HackAttack, line 16: Need to test what the equivalent for Windows is, or if this is even needed on Windows at all.

README.md
```diff
@@ -319,6 +319,12 @@ If there's a git mirror path and `mount-checkout` is enabled, the (mirror path)[
 
 Default: `true`
 
+### `chown` (optional, boolean)
+
+Whether to chown mounted files to the Buildkite agent user before exiting, to ensure that Buildkite can clean them up. Only useful if `mount-checkout` is enabled.
+
+Default: `false`
```
HackAttack, line 326: I think this should default to `true`, since I can’t imagine wanting the prior behavior (which bricks the agent due to inability to e.g. `git clean`), but I made it `false` since it’s a behavior change. If the maintainers are ok with it, happy to change it to `true`, otherwise I recommend that it be flipped in the next major version.

hooks/pre-exit
```diff
@@ -1,10 +1,19 @@
...
 if [[ "${BUILDKITE_PLUGIN_DOCKER_CLEANUP:-true}" =~ ^(true|on|1)$ ]] ; then
   for container in $(docker ps -a -q --filter "label=com.buildkite.job-id=${BUILDKITE_JOB_ID}") ; do
     echo "~~~ Cleaning up left-over container ${container}"
     docker stop "$container"
   done
 fi
 
+if ! is_windows && [[ "${BUILDKITE_PLUGIN_DOCKER_MOUNT_CHECKOUT:-true}" =~ ^(true|on|1)$ ]] && [[ "${BUILDKITE_PLUGIN_DOCKER_CHOWN:-false}" =~ ^(true|on|1)$ ]] ; then
+  docker run --rm -v "$PWD":"$PWD" busybox chown -Rh "$(id -u):$(id -g)" "$PWD"
+fi
```
mcncl, line 18: I don't think we need it, but `chown` in Windows is `cacls`. However given this is new behaviour I think we could consider that later.

mcncl (approved):

README.md
jeremybumsted: I agree `false` makes sense here for now; I think rewording:

> Has no effect on Windows or if `mount-checkout` is false.

to:


> If `mount-checkout` is false, or if using Windows, this option will do nothing.

(or something to that effect) will be help prevent confusion.

hooks/pre-exit
jeremybumsted: it probably makes sense to add it in future, but for now it's fine to just include `chown`

jeremybumsted (approved):

hooks/pre-exit
jeremybumsted: @HackAttack It would also be good to add some tests to make sure this option works as intended, we can help you out with that if you're not sure how :)

README.md
```diff
@@ -287,7 +287,7 @@ Will propagate `GOOGLE_APPLICATION_CREDENTIALS`, `CLOUDSDK_AUTH_CREDENTIAL_FILE_
 
 Whether to match the user ID and group ID for the container user to the user ID and group ID for the host user. It is similar to specifying `user: 1000:1000`, except it avoids hardcoding a particular user/group ID.
 
-Using this option ensures that any files created on shared mounts from within the container will be accessible to the host user. It is otherwise common to accidentally create root-owned files that Buildkite will be unable to remove, since containers by default run as the root user.
+Using this option ensures that any files created on shared mounts from within the container will be accessible to the host user. It is otherwise common to accidentally create root-owned files that Buildkite will be unable to remove—since containers by default run as the root user—unless `chown` is enabled.
```
toote, line 290: I would reword this a bit as there are other considerations to take into account:

```suggestion
Using this option ensures that any files created on shared mounts from within the container will be accessible to the host user. It is otherwise common to accidentally create root-owned files that the agent will be unable to remove, since a lot of container images by default run as the root user. This can also be mitigated for the checkout folder by using the `chown` option of this plugin.
```

README.md
```diff
@@ -319,6 +319,12 @@ If there's a git mirror path and `mount-checkout` is enabled, the (mirror path)[
 
 Default: `true`
 
+### `chown` (optional, boolean)
+
+Whether to chown mounted files to the Buildkite agent user before exiting, to ensure that Buildkite can clean them up. Has no effect on Windows or if `mount-checkout` is false.
```
toote, line 324: I would also add a comment here about this using the `busybox:latest` image as root so that it doesn't come as a surprise that another container is being run. Also, this will run even if the step fails  (because it is in the pre-exit hook) and may take a long time if there are a lot of files in the checkout folder, which may cause issues as there are limits to how long the agent waits until the processes are killed by the agent - specially when the step is cancelled. Also as this only affects the checkout folder and not any other volumes mounted, I would add a comment about the `propagate-uid-gid` option.

Something like:

```suggestion
Whether to `chown` mounted files to the Buildkite agent user before exiting, to ensure that the agent can clean them up.

Note that:
* it will run a `busybox:latest` container image as a root user
* that container will run after the command even if it fails or the step is cancelled
* will only run for your checkout folder (not any other volumes or mounts)
* if your checkout folder is of considerable size, it may add to the execution time
* has no effect on Windows or if `mount-checkout` is false

You may also want to consider using the `propagate-uid-gid` option instead.
```

toote (changes requested):
Very interesting idea. A few comments about all the caveats and warnings for users of this option to take into account and the wording of the documentation.

And agree with other comments about needing tests :)

hooks/pre-exit
HackAttack: Added tests!

README.md
HackAttack: Done.

README.md
HackAttack: Done.

toote (approved):
awesome! :clap:

Issue #279, opened 2024-10-11
Failure to pull docker image does not result in failed buil #109


thiagomata:
Very similar to https://github.com/buildkite-plugins/docker-buildkite-plugin/issues/109

```yaml
- label: ":docker: Isolate issue"
  key: isolate-issue
  agents:
    queue: "default"
  env:
    BUILDKIT_PROGRESS: plain
  plugins:
    - docker#v5.9.0:
        image: "this-image-do-not-exists" # intentionally asking for an image that does not exist
        always-pull: true
        mount-checkout: true
        propagate-environment: true
        privileged: true
        userns: "host"
        shell: false
        volumes:
          - "/var/run/docker.sock:/var/run/docker.sock"
          - "/root/.config/envd:/root/.config/envd"
```

![image](https://github.com/user-attachments/assets/1fe03ef4-1193-4950-bbac-ac648b8b2392)

## Expected Behaviour

The buildkite step should fail.

## Observed Behaviour

The buildkite step concludes as a success.

PR #280, opened 2024-11-06
Fix eval of exit code returned from docker image pull

petetomasik:
Fix eval of returned exit status from `retry()` function.

Addresses: https://github.com/buildkite-plugins/docker-buildkite-plugin/issues/279

### DIFF
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -350,7 +350,6 @@ if [[ "${BUILDKITE_PLUGIN_DOCKER_ALWAYS_PULL:-false}" =~ ^(true|on|1)$ ]] ; then
   echo "--- :docker: Pulling ${image}"
   if ! retry "${BUILDKITE_PLUGIN_DOCKER_PULL_RETRIES:-3}" \
        docker pull "${image}" ; then
-    local retry_exit_status="$?"
     echo "!!! :docker: Pull failed."
     exit "$retry_exit_status"
   fi
--- a/commands/run.sh
+++ b/commands/run.sh
@@ -348,9 +348,8 @@ fi
 
 if [[ "${BUILDKITE_PLUGIN_DOCKER_ALWAYS_PULL:-false}" =~ ^(true|on|1)$ ]] ; then
   echo "--- :docker: Pulling ${image}"
-  retry "${BUILDKITE_PLUGIN_DOCKER_PULL_RETRIES:-3}" docker pull "${image}"
-  retry_exit_status="$?"
-  if [ $retry_exit_status -ne 0 ] ; then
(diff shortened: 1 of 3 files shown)


jeremybumsted (approved):
LGTM

commands/run.sh
```diff
@@ -348,9 +348,9 @@ fi
 
 if [[ "${BUILDKITE_PLUGIN_DOCKER_ALWAYS_PULL:-false}" =~ ^(true|on|1)$ ]] ; then
   echo "--- :docker: Pulling ${image}"
-  if ! retry "${BUILDKITE_PLUGIN_DOCKER_PULL_RETRIES:-3}" \
-       docker pull "${image}" ; then
-    retry_exit_status="$?"
+  retry "${BUILDKITE_PLUGIN_DOCKER_PULL_RETRIES:-3}" docker pull "${image}"
```
toote, line 351: due to the `set -e` at the very top, if the pull fails it will abort the whole script before printing out the error message because you removed the `if`

commands/run.sh
```diff
@@ -348,9 +348,9 @@ fi
...
+  retry_exit_status="$?"
+  if [ $retry_exit_status -ne 0 ] ; then
```
toote, line 353: A `set -e`-friendly way of doing this would be:

```suggestion
  retry "${BUILDKITE_PLUGIN_DOCKER_PULL_RETRIES:-3}" docker pull "${image}" || retry_exit_status="$?"
  if [ ${retry_exit_status:-0} -ne 0 ] ; then
```

toote (changes requested):
In addition to my comment, it would be awesome to add tests (I couldn't find any related to the retry functionality)

commands/run.sh
petetomasik: Thanks for the suggestions!

petetomasik: Added tests for both success and failure conditions on retries of the `docker pull` command.

tests/command.bats
```diff
@@ -66,6 +66,42 @@ setup() {
...
+  export BUILDKITE_PLUGIN_DOCKER_ALWAYS_PULL=true
+  export BUILDKITE_PLUGIN_DOCKER_PULL_RETRIES=2
+
+  stub docker \
+    "pull image:tag" \
+    "pull image:tag"
+
+  run "$PWD"/hooks/command
+
+  assert_failure
```
toote, line 98: I would also test that the exit code is bubbled up to the hook's exit status:

```suggestion
    "pull image:tag : exit 3"

  run "$PWD"/hooks/command

  assert_failure 3
```

toote (approved):
awesome, just a minor comment on the test, but it is mostly cosmetic

pzeballos (approved):
👏🏻

PR #283, opened 2025-05-31
Disable namespace remapping for chown cleanup

HackAttack:
This feature breaks our ability to effectively become root (by design).


toote:
Could you provide more information about the exact issue you are trying to solve? I am not sure I understand what it is.

toote:
Just that the `chown` option will run a container that will share the host's user namespace. Some security people may have issues with that and we want to disclose that kind of things just in case.

HackAttack:
Ok, done.

PR #285, opened 2025-07-04
Adding log driver opts

JoeColeman95:
Adding logging driver and logging options availability to the plugin.
This allows passing of custom `log-driver` and `log-opt`

For example: 

```yml
steps:
  - label: "Hello World"
    command: "echo 'Hello, world'"
    plugins:
      - docker#v5.13.0:
          image: "alpine:latest"
          log-driver: "json-file"
          log-opt:
            - "max-size=1m"
            - "max-file=2"
 ```

This Fixes: #284 

Also updates `README.md` to reflect changes and adds new unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ steps:
   - command: "go build -o dist/my-app ."
     artifact_paths: "./dist/my-app"
     plugins:
-      - docker#v5.13.0:
+      - docker#v5.12.0:
           image: "golang:1.11"
 ```
 
@@ -25,7 +25,7 @@ Windows images are also supported:
 steps:
   - command: "dotnet publish -c Release -o published"
     plugins:
-      - docker#v5.13.0:
+      - docker#v5.12.0:
           image: "microsoft/dotnet:latest"
           always-pull: true
 ```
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -307,6 +323,20 @@ Whether or not to leave the container after the run, or immediately remove it wi
 
 Default: `false`
 
+### `log-driver` (optional, string)
+
+The logging driver for the container. This allows you to configure how Docker handles logs for the container.
+
+Common drivers include: `json-file`, `syslog`, `journald`, `gelf`, `fluentd`, `awslogs`, `splunk`, `etwlogs`, `gcplogs`, `logentries`, `none`.
+
+Default: `json-file`
```
tomowatt, line 332: Aware that the Default log driver is `json-file` for Docker itself, but when specifying the attribute for the Plugin, the _default_ value should represent what the Plugin may use if not set. And in the code below, if nothing is specified, then we do not set the log driver to be used.

Thinking for clarity might be to change this to explain this e.g,

```suggestion
:information_source: As a default, Docker uses the [json-file logging driver](https://docs.docker.com/engine/logging/drivers/json-file/)
```
Maybe?

JoeColeman95: @tomowatt Agreed!

PR #303, opened 2026-03-23
Add reuse-container mode

rsennewald:
Add a persistent container feature. When `reuse-container: true`
is set, the plugin keeps a container alive across job steps
instead of creating and destroying one each time.
Key behaviors:
- Derives a stable container name from the image name and agent
  spawn index (or uses an explicit `reuse-container-name` override).
- On each job, checks for an existing running container: if the
  image digest matches, attaches via `docker exec`; if the digest
  differs, prints a warning banner, removes the stale container,
  and creates a fresh one.
- New containers are started detached (`docker run -d ... sleep
  infinity`) with `--rm` suppressed, then the job command runs via
  `docker exec`.
- Environment variables are stripped from the creation `docker run`
  and injected per-exec only, preventing secrets from leaking
  between jobs that share the container.
- The `pre-exit` hook skips container cleanup when reuse-container
  is enabled so the container persists for subsequent jobs.
Files changed:
- plugin.yml: new `reuse-container` and `reuse-container-name` opts
- lib/shared.bash: `get_reuse_container_name`,
  `get_container_image_id`, `get_image_id` helpers
- commands/run.sh: reuse-container branch with exec-arg extraction,
  digest validation, and detached container creation
- hooks/pre-exit: skip cleanup when reuse-container is set
- tests/reuse-container.bats: 11 bats tests covering creation,
  reuse, digest mismatch, stopped container, custom name, spawn
  index, env passthrough, env stripping, pre-exit, --rm absence,
  and exit code propagation

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,6 +323,20 @@ Whether or not to leave the container after the run, or immediately remove it wi
 
 Default: `false`
 
+### `reuse-container` (optional, boolean)
+
+When set to `true`, the plugin keeps a named container running across pipeline steps instead of creating and destroying one each time. On each step, the plugin checks for an existing container: if the image digest matches, the command runs via `docker exec`; if the digest differs, the old container is removed and a fresh one is created. Environment variables are injected per-exec only (not baked into the container) to prevent secrets from leaking between jobs.
+
+The container name is derived from the image name and the agent's spawn index. If the agent name does not follow the `name-%spawn` convention, set `reuse-container-name` to avoid potential container name collisions between agents on the same host.
+
+Default: `false`
+
+### `reuse-container-name` (optional, string)
+
+Override the auto-generated container name used by the `reuse-container` feature. Use this when the default name derivation does not produce unique names (for example, when multiple agents on the same host do not use the standard `-%spawn` agent name suffix).
+
+Example: `my-build-container`
+
 ### `log-driver` (optional, string)
 
 The logging driver for the container. This allows you to configure how Docker handles logs for the container.
--- a/README.md
+++ b/README.md
@@ -331,6 +331,22 @@ The container name is derived from the image name and the agent's spawn index. I
 
 Default: `false`
 
+#### Reuse safety and lifecycle
+
+A persistent container freezes its bind mounts and other create-time flags, so the plugin guards against reusing one that is no longer correct for the current job:
+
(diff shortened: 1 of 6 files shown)


commands/run.sh
```diff
@@ -591,28 +597,137 @@ elif [[ ${#command[@]} -gt 0 ]] ; then
...
+      -t|-i)
+        exec_args+=("${run_flags[$i]}")
+        ;;
+      --env|--env-file|--workdir|-u)
+        exec_args+=("${run_flags[$i]}" "${run_flags[$((i+1))]}")
+        i=$((i+1))
+        ;;
+    esac
+    i=$((i+1))
+  done
```
toote, line 682: I think it would be more readable to create a separate argument array for the `exec` codepath or, even better, an array for the options available on both `exec` and `run` and another for the ones exclusive to `run`

commands/run.sh
```diff
@@ -591,28 +597,137 @@ elif [[ ${#command[@]} -gt 0 ]] ; then
...
+      windows_multi_command=${BUILDKITE_COMMAND//$'\n'/ && }
+      exec_cmd+=("${windows_multi_command}")
+    else
+      exec_cmd+=("${BUILDKITE_COMMAND}")
+    fi
+  elif [[ ${#command[@]} -gt 0 ]]; then
+    for command_arg in "${command[@]}"; do
+      exec_cmd+=("$command_arg")
+    done
+  fi
```
toote, line 702: same here: why duplicate this logic that is already done above for the `run` command? It would make more sense to save all this to a separate array that is the same for both `run` and `exec`

commands/run.sh
```diff
@@ -591,28 +597,137 @@ elif [[ ${#command[@]} -gt 0 ]] ; then
...
+  # Don't convert paths on gitbash on windows, as that can mangle user paths and cmd options.
+  # See https://github.com/buildkite-plugins/docker-buildkite-plugin/issues/81 for more information.
+  # `trap` is used in this subshell for the same reason it is used above.
+  ( if is_windows ; then export MSYS_NO_PATHCONV=1; fi && trap '' SIGTERM && docker run "${args[@]}" )
+
+  exit_code=$?
+
+  set -e
+
+  exit $exit_code
```
toote, line 841: this whole block is common to both branches, I think it could be simplified with a better crafting (i.e. replacing with a better version) of the  `${args[@]}`

lib/shared.bash
```diff
@@ -76,3 +76,41 @@ function is_macos() {
   [[ "$OSTYPE" =~ ^(darwin) ]]
 }
 
+# Returns a stable container name for the reuse-container feature.
+# Uses the explicit override if set, otherwise derives from the image name
+# and the agent's spawn index (to isolate containers per agent on a host).
```
toote, line 94: this should also be isolated per pipeline, otherwise you could run into issues with a completely different repositories being mounted (if they use the same image to run)

commands/run.sh
toote, line 559: how would would you suggest to fix the fact that the created container will for ever have labels referencing jobs/builds/pipelines it may no longer be running for?

README.md
```diff
@@ -323,6 +323,20 @@ Whether or not to leave the container after the run, or immediately remove it wi
 
 Default: `false`
 
+### `reuse-container` (optional, boolean)
+
+When set to `true`, the plugin keeps a named container running across pipeline steps instead of creating and destroying one each time. On each step, the plugin checks for an existing container: if the image digest matches, the command runs via `docker exec`; if the digest differs, the old container is removed and a fresh one is created. Environment variables are injected per-exec only (not baked into the container) to prevent secrets from leaking between jobs.
```
toote, line 328: you should add a note with a huge warning about turning on this option when running commands that have side-effects on the container or file system as those would be preserved between runs or pipelines

toote (changes requested):
Thanks for the great PR @rsennewald! I think it is a very interesting feature but I have some questions about how it would interact with other parts of the plugin that assume you are running in docker to ensure an ephemeral and isolated execution environment (specially labels and, volume mounts) that is in a known state - because it will be created from an image every time.

Also, I think that the code flow should be better so that we avoid having to duplicate the code to `run` the container or calculate the right arguments for the commands. 

More than happy to discuss these items and looking forward to your opinions!

